PR #38, opened 2018-12-07
Logging cleanup

eanders-MS:
Added a new log facility to enable/disable different categories of logging inside the SDK. Logging can be enabled from an environment variable in a way familiar to node developers: adding the package name to the DEBUG var:
```bash
set DEBUG="@microsoft/mixed-reality-extension-sdk" && node .
```
It can also be set in vscode's launch.json settings.

Logging can be enabled or disabled programmatically as well:
```ts
log.enable(); // Enable all logging
log.enable('network'); // Enable logging only for network activity
log.enable('network', 'errors'); // Only log network errors
log.enable(null, 'errors'); // Log all errors regardless of category
```

Successive calls to `log.enable` are additive. Call `log.disable(...)` to subtract.

#### Other things
- We now log warnings and errors returned in network messages to the console for the developer to see.
- Removed the `Logger` and `Storage` types from the SDK.


eanders-MS:
@tombuMS Correct. I wanted to tackle that separately.

PR #41, opened 2018-12-09
Improved logging to app

eanders-MS:
- Better interface for logging messages to the application.
- Log baseUrl and baseDir at startup.
- Ported functional tests to use WebHost.

### DIFF
--- a/node/packages/sdk/src/webHost.ts
+++ b/node/packages/sdk/src/webHost.ts
@@ -3,7 +3,6 @@
  * Licensed under the MIT License.
  */
 
-import * as http from 'http';
 import * as Restify from 'restify';
 import { Adapter, MultipeerAdapter } from '.';
 import { log } from './log';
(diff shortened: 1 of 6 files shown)


eanders-MS:
lint errors confused as least one of our SDK users this weekend: https://altspacevrsdk.slack.com/archives/C0B39HKLY/p1544329902262400

node/packages/sdk/src/webHost.ts
```diff
@@ -26,33 +26,36 @@ export class WebHost {
         options: { baseDir?: string, baseUrl?: string, port?: string | number }
             = { baseDir: '.', baseUrl: null, port: null }
```
sorenhan: >baseDir: '.' [](start = 16, length = 12)

maybe this is meant to be baseurl: null? that way it would resolvepath to ../public right below, if nothing is passed in?

sorenhan: (Either that, or the samples repo's server.ts files need to be updated, I think)

eanders-MS: Oh good catch, thanks.

PR #43, opened 2018-12-11
Add launch.json for attaching VS Code debugger to the app process.

tombuMS:
Yay for attaching debuggers!


stevenvergenz:
Could you move over the node/.vscode files too? They only work from the root of the project.

tombuMS:
The launch.json in the node/.vscode directory should actually have been the one updated.  Not added to the root.  Eric set this up to open the workspace from the node directory.  I usually open the node directory as well.  VS Code actually added this file from the editor when I added a configuration and hadn't realized that it put it at the root instead of in the node/.vscode folder.

node/.vscode/launch.json
```diff
@@ -1,23 +1,14 @@
...
-  // Use IntelliSense to learn about possible attributes.
-  // Hover to view descriptions of existing attributes.
-  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
-  "version": "0.2.0",
-  "configurations": [
-
-    {
-      "type": "node",
-      "request": "launch",
-      "name": "Launch functional-tests app",
```
eanders-MS: I use this launch task all the time. Why are you removing it?

tombuMS: Must have accidentally stomped this when I was moving files around and merging them together.  Added it back.

PR #76, opened 2018-12-17
Functional Test Browser

sorenhan:
if instantiating functional tests without a parameter then create a menu.

At this point the menu doesn't repopulate after completion, but that appears to be due to a bug somewhere else.

### DIFF
--- a/packages/functional-tests/src/app.ts
+++ b/packages/functional-tests/src/app.ts
@@ -60,16 +60,20 @@ export default class App {
         let testName: string;
         if (Array.isArray(this.params.test) && this.params.test.length > 0) {
             testName = this.params.test[0];
-            await this.startTest(testName);
         } else {
             testName = this.params.test as string;
+        }
+        if (testName) {
+            await this.startTest(testName);
+            this.rpc.send('functional-test:close-connection');
+        } else {
             await this.launchTestBrowser();
         }
     }
-
     private userLeft = (user: MRESDK.User) => {
         console.log(`user-left: ${user.name}, ${user.id}`);
     }
+
 
     private async startTest(testName: string) {
         if (this.activeTests[testName]) {
@@ -83,7 +87,6 @@ export default class App {
             const success = await test.run();
             console.log(`Test complete: '${testName}'. Success: ${success}`);
             this.rpc.send('functional-test:test-complete', testName, success);
-            this.rpc.send('functional-test:close-connection');
 
             delete this.activeTests[testName];
         }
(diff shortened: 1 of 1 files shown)


packages/functional-tests/src/app.ts
```diff
@@ -87,4 +85,56 @@ export default class App {
             delete this.activeTests[testName];
         }
     }
+
+
+    private async launchTestBrowser() {
+        while (true) {
+            const tester = MRESDK.Actor.CreateEmpty(this.context, {});
+            var selectedTestName = undefined;
```
eanders-MS: Did you mean to use `var` here? This should have been flagged by tslint. Also, no need to initialize to `undefined`. It's the default.

sorenhan:
base branch changed. should obviously be for dev

packages/functional-tests/src/app.ts
```diff
@@ -83,4 +85,56 @@ export default class App {
             delete this.activeTests[testName];
         }
     }
+
+
+    private async launchTestBrowser() {
+        while (true) {
```
eanders-MS: Sigh. There must be better solution than `while (true)`. Generators come to mind.

packages/functional-tests/src/app.ts
```diff
@@ -83,4 +85,56 @@ export default class App {
             delete this.activeTests[testName];
         }
     }
+
+
+    private async launchTestBrowser() {
+        while (true) {
+            const tester = MRESDK.Actor.CreateEmpty(this.context, {});
+            var selectedTestName = undefined;
```
eanders-MS: Prefer `let` over `var`, unless you know you need `var`. tslint should have flagged this.

packages/functional-tests/src/app.ts
eanders-MS: I think it's fine actually.

packages/functional-tests/src/app.ts
```diff
@@ -83,4 +85,56 @@ export default class App {
...
     }
+
+
+    private async launchTestBrowser() {
+        while (true) {
+            const tester = MRESDK.Actor.CreateEmpty(this.context, {});
+            var selectedTestName = undefined;
+            await new Promise<void>((resolve) => {
+                let y = 0;
+                for (let key in this.testFactories) {
```
eanders-MS: `for (let key of Object.keys(this.testFactories))`

packages/functional-tests/src/app.ts
```diff
@@ -83,4 +85,56 @@ export default class App {
...
+                            transform: {
+                                position: { x: 0.5, y: y, z: 0 }
+                            }
+                        }
+                    });
+                    y = y + 0.5;
+                }
+            });
+            destroyActors(tester.value);
+            await this.startTest(selectedTestName);
```
eanders-MS: There's a subtle issue here. At the end of `startTest`, it sends an rpc message to the testbed host to close the websocket. This will trigger a `userLeft` event so you'll never see the functional test menu again. Try commenting out this call at the bottom of startTest:
`this.rpc.send('functional-test:close-connection');`

sorenhan: yes, that fixes the test restarting. thank you

packages/functional-tests/src/app.ts
sorenhan: cleaned up

packages/functional-tests/src/app.ts
```diff
@@ -78,9 +84,59 @@ export default class App {
...
 
             delete this.activeTests[testName];
         }
     }
+
+
+    private async launchTestBrowser() {
+        while (true) {
+            const tester = MRESDK.Actor.CreateEmpty(this.context, {});
+            const selectedTestName = await new Promise<string>((resolve) => {
```
eanders-MS: Suggestion: Instead of an inline await on a new Promise, this could just be a call to an async method. Encapsulate the body of this promise into a method.

packages/functional-tests/src/app.ts
```diff
@@ -58,14 +60,18 @@ export default class App {
...
+        } else {
+            await this.launchTestBrowser();
+        }
     }
-
     private userLeft = (user: MRESDK.User) => {
         console.log(`user-left: ${user.name}, ${user.id}`);
     }
 
+
```
eanders-MS: This extra newline should be flagged by tslint.

packages/functional-tests/src/app.ts
```diff
@@ -7,6 +7,8 @@ import * as MRESDK from '@microsoft/mixed-reality-extension-sdk';
 import * as MRERPC from '@microsoft/mixed-reality-extension-sdk/built/rpc';
 import AssetPreloadTest from './tests/asset-preload';
 import ClockSyncTest from './tests/clock-sync-test';
+import delay from './utils/delay';
```
eanders-MS: imports aren't sorted. does tslint not flag this for you?

packages/functional-tests/src/app.ts
```diff
@@ -58,14 +60,18 @@ export default class App {
         } else {
             testName = this.params.test as string;
         }
-
-        await this.startTest(testName);
+        if (testName) {
```
eanders-MS: Did you want to move this logic into onstarted? Seems like the more correct thing to do.

packages/functional-tests/src/app.ts
sorenhan: fixed - lint wasn't enabled by default in this package, which is why I missed it

packages/functional-tests/src/app.ts
sorenhan: fixed

packages/functional-tests/src/app.ts
sorenhan: done

packages/functional-tests/src/app.ts
```diff
@@ -53,37 +55,94 @@ export default class App {
...
+            this.rpc.send('functional-test:close-connection');
+        } else {
+            if (!this.firstUser) {
+                this.runFunctionalTestMenu();
+            }
+            this.firstUser = user;
+        }
     }
-
     private userLeft = (user: MRESDK.User) => {
```
eanders-MS: If the last user leaves then another one joins, I think it will be broken. This means the app won’t be able to run in a long-lived way. You could handle this by clearing `firstUser` in `userLeft` if the user.id matches.

eanders-MS: Oh wait, I see you're overwriting firstUser. nevermind!

PR #80, opened 2018-12-18
changes needed to deploy functional tests to openode

sorenhan:
This is what I needed to changed for the openode.io deploy to just work for functional tests.

I couldn't explain why tslib is needed and our local builds don't complain about it, so I'm adding in the reference for now and an issue to track it whenever we need to debug it


eanders-MS:
I agree with adding tslib to the dependencies. That should be sufficient to get it deploying. If we think we want to split functional-tests from the monorepo, then we could do something like what Tom did for the samples.. add a script to toggle the SDK dependency from npm to local when needed.

sorenhan:
I definitely do not want us to split the functional tests from the repo. It's super convenient to have in there

PR #84, opened 2018-12-20
Merge dev to master (minus commit 71bf2a5)

eanders-MS:
This is everything from dev except my lookAt changes in commit #71bf2a5


eanders-MS:
Tests with the current release of Altspace. LookAt was functional, so I think I'm good to merge this.

PR #86, opened 2018-12-21
WebHost buffer hosting, gltf-gen spheres

stevenvergenz:
* Add the `registerStaticBuffer` method to WebHost, allowing users to serve buffers through the web server
* Add sphere geometry prims to gltf-gen
* Add a gltf-gen functional test


packages/functional-tests/src/app.ts
```diff
@@ -40,6 +41,7 @@ export default class App {
         'clock-sync-test': (): Test => new ClockSyncTest(this, this.baseUrl),
         'primitives-test': (): Test => new PrimitivesTest(this, this.baseUrl),
         'input-test': (): Test => new InputTest(this, this.baseUrl),
+        'gltf-gen': (): Test => new GltfGen(this, this.baseUrl),
```
eanders-MS: By convention, test name should end in '-test'. --> 'gltf-gen-test'

packages/functional-tests/src/tests/gltf-gen.ts
```diff
@@ -0,0 +1,48 @@
...
+import { resolve } from 'path';
+
+export default class GltfGenTest extends Test {
+
+    constructor(app: App, private baseUrl: string) {
+        super(app);
+    }
+
+    public async run(): Promise<boolean> {
+        let spherePrim = new GltfGen.Sphere(0.5);
```
eanders-MS: Use const here. Did tslint not flag this? Wondering if lint is not working for you on the functional-tests project.

packages/functional-tests/src/tests/gltf-gen.ts
```diff
@@ -0,0 +1,48 @@
...
+        spherePrim.material = new GltfGen.Material({
+            baseColorTexture: new GltfGen.Texture({
+                source: new GltfGen.Image({
+                    embeddedFilePath: resolve(__dirname, '../../public/uv-grid.png')
+                })
+            })
+        });
+        let gltfFactory = new GltfGen.GltfFactory([new GltfGen.Scene({
```
eanders-MS: Also use const here. #tslint

packages/functional-tests/src/app.ts
```diff
@@ -85,7 +87,14 @@ export default class App {
             const test = this.activeTests[testName] = this.testFactories[testName](user);
             this.rpc.send('functional-test:test-started', testName);
             console.log(`Test started: '${testName}'`);
-            const success = await test.run();
+            let success: boolean;
+            try {
+                success = await test.run();
+            }
+            catch (e) {
```
eanders-MS: catch should appear on same line as `}` --> `} catch (e) {`. I think tslint isn't working for you in this project.

packages/functional-tests/src/app.ts
eanders-MS: Your addition of gltf-gen is different then every other test. Let's stick with one convention, for discoverability's sake.

eanders-MS: Agreed it is redundant. Alternatively, could remove '-test' from all of them.

sorenhan: While it is redundant, I personally really like that it's there. I love that I can see on the filename if it's a test file (for example text-test.ts vs text.ts). Similarly regarding callstack/autocomplete/find in files, I don't care too much about the string name, though. So I'd rather we kept the word test on all of them

sorenhan:
You opened this PR to pull it into Master, but it's supposed to be dev, until we update our workflow.

packages/functional-tests/src/app.ts
stevenvergenz: It just bugs me that I have to type the word "test" three times when I write up the URL haha. I'll add it.

stevenvergenz:
Closing and reopening against dev

PR #92, opened 2018-12-21
Stvergen/user property bag

stevenvergenz:


packages/sdk/src/types/runtime/user.ts
```diff
@@ -16,6 +16,11 @@ export interface UserLike {
     name: string;
     transform: TransformLike;
     subscriptions: SubscriptionType[];
+
+    /**
+     * A grab bag of miscellaneous, possibly host-dependent, properties.
+     */
+    properties: { [name: string]: any };
```
stevenvergenz: Can we document the `remoteAddress` property here? What would that syntax look like?

packages/sdk/src/protocols/execution.ts
```diff
@@ -4,7 +4,7 @@
  */
 
 import { Protocol, ServerPreprocessing } from '.';
-import { ActionEvent, CollisionEvent, Context } from '..';
+import { ActionEvent, CollisionEvent, Context, WebSocket } from '..';
```
eanders-MS: This is not a safe assumption. Not all connections are WebSockets.

stevenvergenz: No assumption is made, this is just an import statement. Further down in both adapters, I ask `if (this.conn instanceof WebSocket)`. Since multipeer connects to sessions via the WebSocket adapter, it all irons out in the end.

PR #94, opened 2018-12-22
Release 2018.12.21

stevenvergenz:
I'm not willing to add a triple and quadruple revert to dev, so here we are. This PR merges dev (minus 07f1f2c) into master. Once it's done I can push a new npm package.


packages/functional-tests/src/app.ts
```diff
@@ -32,16 +34,18 @@ export default class App {
...
         'rigid-body-test': (): Test => new RigidBodyTest(this),
         'text-test': (): Test => new TextTest(this),
         'clock-sync-test': (): Test => new ClockSyncTest(this, this.baseUrl),
+        'library-test': (): Test => new LibraryTest(this, this.baseUrl),
         'primitives-test': (): Test => new PrimitivesTest(this, this.baseUrl),
         'input-test': (): Test => new InputTest(this, this.baseUrl),
+        'gltf-gen-test': (): Test => new GltfGenTest(this, this.baseUrl),
         'root-motion-test': (): Test => new RootMotionTest(this, this.baseUrl),
-        'asset-preload-test': (user: MRESDK.User): Test => new AssetPreloadTest(this, this.baseUrl, user)
+        'asset-preload': (): Test => new AssetPreloadTest(this, this.baseUrl)
```
eanders-MS: Looks like a regression here. should be `asset-preload-test`

packages/functional-tests/src/app.ts
```diff
@@ -76,16 +80,22 @@ export default class App {
...
     }
-    private async startTest(testName: string,  user: MRESDK.User) {
+    private async startTest(testName: string, user: MRESDK.User) {
         if (this.activeTests[testName]) {
             console.log(`Test already running: '${testName}'`);
         } else if (!this.testFactories[testName]) {
             console.log(`error: Unrecognized test: '${testName}'`);
         } else {
-            const test = this.activeTests[testName] = this.testFactories[testName](user);
+            const test = this.activeTests[testName] = this.testFactories[testName]();
```
eanders-MS: Did you lose the `user` parameter here?

packages/functional-tests/src/tests/asset-preload.ts
```diff
@@ -11,7 +11,7 @@ import Test from './test';
 
 export default class AssetPreloadTest extends Test {
 
-    constructor(app: App, private baseUrl: string, private user: MRESDK.User) {
+    constructor(app: App, private baseUrl: string) {
```
eanders-MS: Lost the user parameter here?

packages/functional-tests/src/app.ts
sorenhan: no, this was actually not a regression ,this was because the fixes was in the revert-revert-revert. Don't worry about this one

packages/functional-tests/src/app.ts
eanders-MS: Oh I think this is correct.

sorenhan: that's because it's part of the commit being reverted. it was stuff that came as part of your commit that is being reverted

packages/functional-tests/src/tests/asset-preload.ts
eanders-MS: Also correct, I think

packages/functional-tests/src/tests/asset-preload.ts
```diff
@@ -11,7 +11,7 @@ import Test from './test';
 
 export default class AssetPreloadTest extends Test {
```
eanders-MS: This filename should be asset-preload-test.ts

packages/functional-tests/src/tests/asset-preload.ts
stevenvergenz: See above.

packages/functional-tests/src/app.ts
stevenvergenz: Good catch!

packages/sdk/src/webHost.ts
```diff
@@ -57,10 +61,41 @@ export class WebHost {
...
+        const matches = this.bufferRegex.exec(req.url);
+        const procPath = matches && matches[1] || null;
+
+        // see if there's a handler registered for it
+        if (!procPath || !this.bufferMap[procPath]) {
+            return next();
+        }
+
+        // if so, serve binary
+        res.sendRaw(200, this.bufferMap[procPath]);
```
eanders-MS: Oh missed this in the other PR, it is more correct to always call next(), even after res.send.

PR #106, opened 2019-01-11
Soren/exclude openode

sorenhan:
to stop accidental check-ins of .openode config files


eanders-MS:
Did you mean to pull in gltf-concurrency-test too?

sorenhan:
@eanders-MS I had the wrong target branch - changing to dev fixed this.

PR #108, opened 2019-01-12
[SDK] Animation system updates

eanders-MS:
*Merging to `awaiting` branch because this change requires a DLL update*

#### Summary
- Simplified the animation API. Client remains backward compatible with the older API (except for root motion, see below), so I think we're fine to leave the protocol version where it is.
- Added `Actor.animateTo` method. It supports ease curves.
- Removed root motion support in favor of `animateTo`. Soren and I discussed.
- Added 'ping-pong' animation wrap mode.
- Added 'interpolation-test' functional test.

#### `animateTo` Example
```ts
// Move the object upward on an ease/out curve over one second
actor.animateTo({ transform: { position: { y: 5 } } }, 1.0, AnimationEaseCurves.EaseInOutSine);
```

**NOTE: MERGE BLOCKED ON NEXT RELEASE OF ALTSPACE CLIENT**
The `awaiting` branch is locked until after the next AltspaceVR release.


AndreMuis:
Looks like now all animations are ease in / ease out. Now we can specify. Cool.

packages/functional-tests/src/tests/asset-preload-test.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Licensed under the MIT License.
+ */
+
+import * as MRESDK from '@microsoft/mixed-reality-extension-sdk';
+import App from '../app';
+import delay from '../utils/delay';
+import destroyActors from '../utils/destroyActors';
+import Test from './test';
+
+export default class AssetPreloadTest extends Test {
```
stevenvergenz: How is this a new file? This is an old test.

packages/functional-tests/src/tests/root-motion-test.ts
```diff
@@ -1,190 +0,0 @@
-/*!
- * Copyright (c) Microsoft Corporation. All rights reserved.
- * Licensed under the MIT License.
- */
-
-import * as MRESDK from '@microsoft/mixed-reality-extension-sdk';
-import { Actor, AnimationWrapMode } from '@microsoft/mixed-reality-extension-sdk';
-import { promises } from 'fs';
```
stevenvergenz: Is this really deleted?

packages/sdk/src/types/internal/actor.ts
```diff
@@ -33,6 +33,18 @@ export class InternalActor {
         }
     }
 
+    public setAnimationStateEventRaised(animationName: string, state: SetAnimationStateOptions) {
+        if (this.actor) {
+            if (state.enabled !== undefined) {
```
stevenvergenz: Combine conditions

stevenvergenz:
Is root motion not a thing anymore? It seems to have been removed from the API.

packages/functional-tests/src/tests/asset-preload-test.ts
eanders-MS: Updated the file name, since it was the only one that didn't end in `-test`. It seems someone else did the same.

packages/functional-tests/src/tests/root-motion-test.ts
eanders-MS: Yes (as we discussed offline, subsumed by `animateTo`)

packages/sdk/src/types/internal/actor.ts
eanders-MS: Leaving this for now. There may be other sub-conditions coming later.

PR #109, opened 2019-01-12
Respect proxy/forwarder headers when determining the connection's source

willneedit:
Currently, the remote address passed to the client's user.properties.remoteAddress is just the address of the last hop connecting to the MRE.

This fails whenever a proxy forwarded this request, e.g. Heroku's load balancers - the IP address shown would be the outgoing interface of the load balancer.

This change includes 'forwarded-for' which scans the HTTP headers for 'well-known' forwarder header lines like 'x-forwarded-for' and adjusts the IP address to be passed on accordingly. Without a proxy inbetween, forwarded-for would take the original remoteAddress of the request.

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -47,6 +47,7 @@
     "@types/node": "^10.3.1",
     "debug": "^4.1.0",
     "deepmerge": "^2.1.1",
+    "forwarded-for": "^1.0.1",
     "query-string": "^6.2.0",
     "restify": "^7.2.0",
     "uuid": "^3.2.1",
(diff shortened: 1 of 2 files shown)


eanders-MS:
Good catch. Only this PR needs to target the dev branch (that's where changes go first... usually). Looks good otherwise. We're working to document our branch structure.

The 'forwarded-for' package is MIT-licensed and has no additional dependencies that may complicate things, so we're ok taking a dependency on it.

eanders-MS:
Yes please rebase to dev. Thanks.

PR #112, opened 2019-01-14
[SDK] Materials support (immutable)

stevenvergenz:
* Strip down the Material asset type to what will actually be used
* Add `Actor.materialId` and `Actor.material` properties
* Modify asset-preload test to exercise material load/set/unset


packages/sdk/src/types/runtime/actor.ts
```diff
@@ -104,14 +109,29 @@ export class Actor implements ActorLike {
...
         if (!this.context.actor(value)) {
-            value = undefined; // throw?
+            value = null; // throw?
         }
         this._parentId = value;
         this.actorChanged('parentId');
     }
+    public get material() { return this._context.assets.byId(this._materialId) as Material; }
+    public set material(value) {
+        this.materialId = value && value.id || undefined;
```
eanders-MS: `null` here as well?

stevenvergenz: Good catch!

packages/sdk/src/types/runtime/assets/material.ts
```diff
@@ -11,23 +11,7 @@ import { Color3, Color4 } from '../../../math';
  */
 export interface MaterialLike {
     /** The base color of this material. */
-    color: Color4;
-    /** The texture ID assigned to this material, if any. */
-    textureId?: string;
-    /**
```
eanders-MS: Would be great to support metallic and roughness, at least. Is this still planned?

stevenvergenz: Not technically. Our shader is vertex-lit, so reflection parameters are entirely unused. Don't want to get users excited over nothing.

eanders-MS:
Wait, I thought we agreed to hold off on this merge.

eanders-MS:
After Altspace vNext ships, awaiting will be merged to master and a new npm package published. We don't want these changes to go out in that package.

PR #121, opened 2019-01-18
[SDK] Asset mutability

stevenvergenz:
* Updated `Material` to generate patches.
* Implemented interfaces `Patchable<T>` and `InternalPatchable<T>`, which the context uses to check for updates.
* `StateUpdate` messages can now contain both `ActorUpdate`s and `AssetUpdate`s.
* Added a new functional test that demonstrates shared mutable materials.
* Renamed the context asset manager to `assetManager` from `assets`. Also rearranged some APIs in the manager for better usability.

### DIFF
--- a/packages/sdk/src/types/runtime/assets/material.ts
+++ b/packages/sdk/src/types/runtime/assets/material.ts
@@ -61,6 +61,7 @@ export class Material extends Asset implements MaterialLike, Patchable<AssetLike
         }
 
         this._color.copy(def.material.color);
+        // material patching: observe the color for changed values, and write them to a patch
         observe(this._color, 'color', (...path: string[]) => this.materialChanged(path));
     }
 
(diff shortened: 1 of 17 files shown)


packages/sdk/src/types/runtime/assets/prefab.ts
```diff
@@ -4,17 +4,23 @@
...
+import { InternalAsset } from '../../internal/asset';
+import { Patchable } from '../../patchable';
 
 export interface PrefabLike {
     /** The number of actors this prefab contains. */
     actorCount: number;
 }
 
-export class Prefab extends Asset implements PrefabLike {
+export class Prefab extends Asset implements PrefabLike, Patchable<AssetLike> {
```
eanders-MS: Are prefabs going to be patchable?

packages/sdk/src/types/runtime/assets/material.ts
```diff
@@ -53,10 +61,26 @@ export class Material extends Asset implements MaterialLike {
         }
 
         this._color.copy(def.material.color);
+        observe(this._color, 'color', (...path: string[]) => this.materialChanged(path));
```
eanders-MS: Please put a comment above this line similar to the ones you'll find above `observe` calls in the Actor class. The patching system is still pretty cryptic. Explanatory comments will help future devs.

PR #125, opened 2019-01-19
Fix serialization failure when nulling synced actor property

stevenvergenz:
This function threw an exception if `null` were passed in, as its type is `'object'`, but doesn't support Object.keys.

### DIFF
--- a/packages/sdk/src/utils/filterEmpty.ts
+++ b/packages/sdk/src/utils/filterEmpty.ts
@@ -8,7 +8,9 @@
  * If `obj` is an empty object, return undefined.
  */
 export default function filterEmpty(obj: any) {
-    if (typeof obj !== 'object' || Object.keys(obj).length) {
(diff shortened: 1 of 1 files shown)


packages/sdk/src/utils/filterEmpty.ts
```diff
@@ -8,7 +8,9 @@
  * If `obj` is an empty object, return undefined.
  */
 export default function filterEmpty(obj: any) {
-    if (typeof obj !== 'object' || Object.keys(obj).length) {
+    if (typeof obj === 'object' && obj !== null && !Object.keys(obj).length) {
```
eanders-MS: This should be a less strict check, to catch `undefined`.

stevenvergenz: Undefined is not of type object, so this PR won't have affected that path. It would `return obj`, which in this case is the same thing as `return undefined`.

eanders-MS: Ah, true.

Issue #124, opened 2019-01-19
Text is backward


stevenvergenz:
Text faces the wrong way: -Z instead of the Unity conventional +Z.

sorenhan:
We can either fix it, or provide helper functions for the cases where it hurts, or choose to live with this. There is a cost (both code complexity and performance) with simply flipping the text and leaving everything the same)

stevenvergenz:
Agreed, except for that I don't know of a way to apply a transform to it without generating a new GameObject. And that introduces a lot of complexity too.

sorenhan:
Text flipping investigation:
1. Testbed/TextMesh: We currently don't have a custom shader in our unity samples. We actually have a z-sorting shader in the testbed unity scene, and use it for all the custom testbed elements, we just don't use it in the actual SDK elements. We should be able to switch that, and make the flip-on-x change in there. Sounds pretty easy.
2. AltspaceVR/TextMeshPro: The shader is available, and we can modify it, and it mostly worked when doing vert.x*=1.0f; - it does look like it's just one drawcall with all letters being added relative to an offset. There were problems:
a. it would affect altspace elements, too -> we'd need to fix things
b. culling needs to also be updated, that's outside the shader
c. it still does some weird jumping in some instances, as if the origin can move around, and the shader counteracts
d. sometimes, but not always, text changed to purple squares. 
 
I see 3 possible solutions:
1. add extra child gameObject for the text issues, with x=-1. Extra gameObject overhead
2. modify shaders, and compensate in TextMeshPro as needed. Custom TMP shader=overhead when merging in new TMP changes, and possibly a per-pixel overhead. Not sure how extensive the TMP changes would be - it may be easy and reliable, or may end up being fragile, more investigation needed
3. Accept that text is backwards, and add a LookAwayFrom option to the lookAt functionality

sorenhan:
We've decided that "text is backward" is the a perfectly fine solution, so we won't even try to change it

PR #126, opened 2019-01-19
Update text orientation for all functional tests

stevenvergenz:
Rotate all the text 180 degrees to compensate for #124


tombuMS:
I agree. If this is an issue specific to how unity handles the z direction, it should be fixed in our Unity code wouldn't you agree?

stevenvergenz:
I agree, except I don't know how to! There's no transform field on text, and creating a new GameObject to rotate it adds a lot of complexity. Any preferred strategies for flipping it?

eanders-MS:
Is it possible to reverse the UV's on the text texture?

sorenhan:
As we talked about, we will hold off on this until we have a solution to flip the text in both TextMesh and TextMeshPro

PR #143, opened 2019-01-26
Reject loadGltf promise if failureMessage returned

stevenvergenz:


packages/sdk/src/types/runtime/assets/assetManager.ts
```diff
@@ -85,6 +86,9 @@ export class AssetManager {
         } as LoadAssets;
 
         const response = await this.sendLoadAssetsPayload(payload);
+        if (response.failureMessage) {
+            throw new Error(response.failureMessage);
+        }
```
eanders-MS: This needs to be converted to a rejected promise in the calling method.

stevenvergenz: It's in an async method, it's already a promise rigged to reject on errors, right?

eanders-MS: Oh yes! Missed that detail.

---
In reply to: [251176960](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/143#discussion_r251176960) [](ancestors = 251176960)

Issue #110, opened 2019-01-14
AssetManager.loadGltf results in unexpected reply message


eanders-MS:
It seems the reply message isn't getting routed to the pending promise.

eanders-MS:
Fixed by #147

Issue #141, opened 2019-01-25
Material changes aren't fully synchronized


sorenhan:
Asset preload test doesn't synchronize all material changes reliably

eanders-MS:
Fixed by #147

Issue #130, opened 2019-01-23
Behaviors: App receives duplicate input events (hover enter, exit, etc)


eanders-MS:
This is happening in the client join synchronization process. The set-behavior message is getting sent twice.

eanders-MS:
Fixed by #147

PR #147, opened 2019-01-28
Big cleanup of message handling in the multi-peer adapter

eanders-MS:
A clear pattern finally emerged in the way the multi-peer adapter handle's different kinds of network messages at various stages of handshake, join-in-progress, and regular operation.

This refactor formalizes that pattern, standardizing how message handling is configured for every payload type at different join stages. Comprehensive message handling "rules" are defined in a new file called 'rules.ts'.

The code is still complex, but hopefully more understandable and maintainable now.

Fixes: #110, #141, #130, the bug where the actor transform was always synced, and probably others.

This is a big PR. Review what you can.

### DIFF
--- a/packages/sdk/src/adapters/multipeer/protocols/clientSync.ts
+++ b/packages/sdk/src/adapters/multipeer/protocols/clientSync.ts
@@ -3,20 +3,43 @@
  * Licensed under the MIT License.
  */
 
-import { Client, SyncActor } from '..';
-import * as MRESDK from '../../..';
+import { Client, MissingRule, Rules, SyncActor } from '..';
+import { Message } from '../../..';
+import { log } from '../../../log';
 import * as Protocols from '../../../protocols';
 import * as Payloads from '../../../types/network/payloads';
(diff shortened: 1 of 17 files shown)


packages/sdk/src/adapters/multipeer/protocols/clientSync.ts
```diff
@@ -25,109 +48,168 @@ export class ClientSync extends Protocols.Protocol {
...
-    private createBehaviors(): void {
-        for (const actor of this.client.session.actors) {
-            this.createActorBehavior(actor);
-        }
+    /**
+     * @hidden
+     * Driver for the `create-animations` synchronization stage.
+     */
+    public 'stage:create-animations' = async () => {
+        // Send all cached interpolate-actor and creaet-animation messages.
```
eanders-MS: >creaet [](start = 49, length = 6)

speling

eanders-MS:
Although this is an SDK-only change, given the size of this change I'm retargeting to awaiting to allow more bake time.

PR #151, opened 2019-01-29
display on-screen error if functional test fails/throws exception. 

sorenhan:
and clean up event handlers on context exit, clean up actors and event handlers on test complete

### DIFF
--- a/packages/altspacevr-extras/src/videoPlayerManager.ts
+++ b/packages/altspacevr-extras/src/videoPlayerManager.ts
@@ -54,7 +54,6 @@ export class VideoPlayerManager {
                         });
                 }
             }
-            userRPC.cleanup();
         }
         this.hasAnyUserJoined = true;
     }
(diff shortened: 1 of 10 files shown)


packages/sdk/src/types/runtime/context.ts
```diff
@@ -112,11 +121,20 @@ export class Context {
...
+        this.emitter.off('user-left', handler);
+        return this;
+    }
+
     /**
      * @hidden
      * (for now)
      */
-    public onActorCreated(handler: (user: User) => void): this {
+    public onActorCreated(handler: (actor: Actor) => void): this {
```
eanders-MS: Haha good catch.

packages/altspacevr-extras/src/videoPlayerManager.ts
```diff
@@ -47,6 +54,7 @@ export class VideoPlayerManager {
                         });
                 }
             }
+            userRPC.cleanup();
```
eanders-MS: It isn't clear to me why you need to do this here. Add a comment?

PR #155, opened 2019-01-30
Improvements to client join process

### LINKED ISSUE #104: Stale connections pinging the MRE, causing a deadlock
On running the Stargate in a larger scale, I noticed that I got connection attempts from two distinct IP addresses which spawned the MRE's in an otherwise empty space.

Biggest problem is that since these 'stale' instance become authoritative for the geometry and animations displayed, they prevent the MRE from functioning for every other user at all.

Resetting the space might get rid of it, and (untested) blocking the IP's at link level might mitigate the problem.

Additionally, Heroku cuts idle connections after 55 seconds, but the logs show these IP addresses over and over again, so they definitely reconnect.

eanders-MS:
This changeset started out as an update to the user join process to add protections against zombie sessions. Along the way I encountered and fixed a few other issues (list below).

**Glossary**
*Zombie session*: A session where the authoritative peer has become unresponsive but not disconnected.

Zombie session mitigations:

* **Handshake new connections before adding to session**: When a new WebSocket connection is made we no longer spawn an app session right away. We now run through the handshake process first. Once handshake completes we pass the connection to a session where join-in-progress is initiated. This should filter out the unresponsive new connections reported by @willneedit.

* **Kick unresponsive peers**: Messages sent to the client that expect replies now have timeouts on them. Default value is 30 seconds. This can be disabled by setting ConnectionTimeoutSeconds to zero (helpful when sitting on a breakpoint). When a message timeout triggers, the connection is forcibly closed.

Fixes #104 

Other changes:
- Only forward `assets-loaded` messages to the app if sent from the authoritative peer.
- When a connection closes unexpectedly, reject outstanding promises against sent messages that are awaiting a reply.
- When enabling rigid body, don't automatically subscribe to rigid body changes.
- Removed `position` and `rotation` from `RigidBody` class.
- When enabling a rigid body, text, or light component, don't send default values for fields not specified by the user. For example, if the user didn't specify an angular velocity, don't pass [0,0,0]. Omit the field instead.
    - Default values for all fields will come back in the next actor-update.
- Programmatically sort the functional test menu.
- Companion fix in Unity for broken animation synchronization.
- Better error handling throughout.


packages/functional-tests/src/app.ts
```diff
@@ -116,44 +116,42 @@ export default class App {
     private async displayFunctionalTestChoices(rootActor: MRESDK.ActorLike): Promise<string> {
         return new Promise<string>((resolve) => {
             let y = 0;
-            for (const key in this.testFactories) {
-                if (this.testFactories.hasOwnProperty(key)) {
```
sorenhan: >this.testFactories.hasOwnProperty(key) [](start = 20, length = 38)

I thought this was required, TS-lint complaining if missing?

eanders-MS: It's required with for..in, but not for..of.

packages/sdk/src/adapters/multipeer/session.ts
```diff
@@ -68,79 +69,87 @@ export class Session extends EventEmitter {
...
-        this._conn.off('close', this.disconnect);
-        this._conn.off('error', this.disconnect);
-        this._conn.close();
-        this.emit('close');
+        try {
+            this._conn.off('close', this.disconnect);
+            this._conn.off('error', this.disconnect);
+            this._conn.close();
+            this.emit('close');
+        } catch { }
```
sorenhan: >catch { } [](start = 10, length = 9)

should you log error here?

eanders-MS: I'm intentionally not logging errors here because we're in the process of tearing down the session. Various errors can happen depending on how "started up" the session was in the first place. I just want to make sure exceptions don't interrupt the shutdown process from completing.

---
In reply to: [252832190](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/155#discussion_r252832190) [](ancestors = 252832190)

Issue #149, opened 2019-01-29
Remove BufferedEventEmitter


sorenhan:
Right now we buffer various events (onactorcreated/onactordestroyed/onuserjoined etc), until an event handler is added to listen. This is problematic because long-lived contexts will just keep allocating memory over time. The right solution is to not buffer these events, and instead code that adds event handlers (for example onUserJoined) should be able to execute that event handler on a list of valid objects (for a userJoined callback it would be on all already-joined users)

PR #156, opened 2019-01-30
Remove BufferedEventEmitter

eanders-MS:
Turns out we don't need BufferedEventEmitter, and it was accumulating events for unassigned handlers, eating up memory.

Fixes issue #149


sorenhan:
Approved for RED (again, could go to green, but red is probably better unless there is an explicit reason to do it faster)

PR #157, opened 2019-01-31
[SDK] New version compatibility scheme

eanders-MS:
Soren and I worked out a new version checking scheme: Client reports current client version and minimum supported SDK version. App checks this against minimum supported client version and current SDK version. If current client version >= minimum supported client version && current SDK version >= minimum supported SDK version, the connection is allowed.

### DIFF
--- a/packages/sdk/src/utils/verifyClient.ts
+++ b/packages/sdk/src/utils/verifyClient.ts
@@ -3,15 +3,24 @@
  * Licensed under the MIT License.
  */
 
+import semver from 'semver';
 import * as Constants from '../constants';
+import { log } from '../log';
 
 /*
- * Supported protocol version
- * **WARNING**: Incrementing this value will break older clients. Strive to be backward compatible. Change this value
- * only after discussing with the team and considering every alternative.
+ * Current SDK Version - Read from package.json.
+ */
+// tslint:disable-next-line:no-var-requires variable-name
+const CurrentSDKVersion = semver.coerce(require('../../package.json').version);
+
+/*
+ * Minimum Supported Client Library version
+ * As part of the manual SDK release procedures, this is reset to match CurrentSDKVersion Major.Minor if new functions
+ * have been added that don't work on older clients (i.e. pretty much every release). Since host apps are required to
+ * update client libraries regularly, this one is not a big deal to update.
  */
 // tslint:disable-next-line:variable-name
-const CurrentProtocolVersion = 1;
+const MinimumSupportedClientVersion = semver.coerce('0.5');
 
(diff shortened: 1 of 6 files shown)


packages/sdk/src/utils/verifyClient.ts
```diff
@@ -3,61 +3,83 @@
...
+
+/*
+ * Minimum Supported Client Library version
+ * As part of the manual SDK release procedures, this is reset to match CurrentSDKVersion Major.Minor if new functions
+ * have been added that don't work on older clients (i.e. pretty much every release). Since host apps are required to
+ * update client libraries regularly, this one is not a big deal to update.
  */
 // tslint:disable-next-line:variable-name
-const CurrentProtocolVersion = 1;
+const MinimumSupportedClientVersion = semver.coerce('0.5');
```
stevenvergenz: Does this mean that client versions are expected to track their SDK versions?

packages/sdk/src/utils/verifyClient.ts
```diff
@@ -3,61 +3,83 @@
...
+
+    // Test the current client version. Is it greater than or equal to the minimum client version?
+    const clientPass = semver.gte(CurrentClientVersion, MinimumSupportedClientVersion);
+    // Test the current SDK version. Is it greater than or equal to the minimum SDK version?
+    const sdkPass = semver.gte(CurrentSDKVersion, MinimumSupportedSDKVersion);
+
+    if (!clientPass) {
+        // tslint:disable-next-line:max-line-length
+        const message = `Connection rejected due to out of date client. Client version: ${CurrentClientVersion.toString()}. Min supported version by SDK: ${MinimumSupportedClientVersion.toString()}`;
+        log.info('network', message);
```
stevenvergenz: >log.info('network' [](start = 8, length = 18)

I think these messages should be exposed to the app developer by default, not hidden with network traffic. Also, probably an `error` or `warning`, not an `info` message.

packages/sdk/src/utils/verifyClient.ts
eanders-MS: The client needs to track the minimum SDK version it supports, but it shouldn't change often. We must make efforts to remain as backward compatible as possible. Incrementing this version in the client is a Big Deal, and requires team discussion and signoff.

packages/sdk/src/utils/verifyClient.ts
eanders-MS: True, they shouldn't be in with the network traffic.

sorenhan:
approved for RED branch

PR #164, opened 2019-02-01
[SDK] Texture asset support

stevenvergenz:
* Instrument Vector2 to be patchable
* Add a ZeroGuid constant
* FIll out the Texture asset class
* Add a texture, offset, and scale to Materials
* Update functional tests to use new features

### DIFF
--- a/packages/sdk/src/math/vector2.ts
+++ b/packages/sdk/src/math/vector2.ts
@@ -7,10 +7,15 @@ import { Epsilon, Matrix, Scalar, Vector3 } from '.';
 
 // tslint:disable:member-ordering variable-name one-variable-per-declaration trailing-comma no-bitwise curly max-line-length
 
+export interface Vector2Like {
+    x: number;
+    y: number;
+}
+
 /**
  * Class representing a vector containing 2 coordinates
  */
-export class Vector2 {
+export class Vector2 implements Vector2Like {
 
     // Statics
 
@@ -655,5 +660,16 @@ export class Vector2 {
      */
     public clone(): Vector2 {
         return new Vector2(this.x, this.y);
+    }
+
+    /**
+     * Updates the Vector2 from the sparsely populated value.
+     * @param from The sparsely populated value to read from.
+     */
+    public copy(from: Partial<Vector2Like>): this {
+        if (!from) { return this; }
+        if (from.x !== undefined) { this.x = from.x; }
+        if (from.y !== undefined) { this.y = from.y; }
+        return this;
     }
 }
--- a/packages/sdk/src/math/vector2.ts
+++ b/packages/sdk/src/math/vector2.ts
@@ -305,7 +305,7 @@ export class Vector2 implements Vector2Like {
         return {
             x: this.x,
             y: this.y,
-        };
+        } as Vector2Like;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


packages/sdk/src/math/vector2.ts
```diff
@@ -7,10 +7,15 @@ import { Epsilon, Matrix, Scalar, Vector3 } from '.';
...
+export interface Vector2Like {
+    x: number;
+    y: number;
+}
+
 /**
  * Class representing a vector containing 2 coordinates
  */
-export class Vector2 {
+export class Vector2 implements Vector2Like {
```
eanders-MS: >Vector2 [](start = 13, length = 7)

This class also needs a toJSON() method. Copy from Vector3 along with the comment above it.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -27,6 +27,7 @@ import {
     SetAnimationStateOptions
 } from '../..';
 import { log } from '../../log';
+import { ZeroGuid } from '../../utils/deterministicGuids';
```
eanders-MS: >ZeroGuid [](start = 9, length = 8)

constants.ts would be a better place for this.

packages/sdk/src/utils/deterministicGuids.ts
```diff
@@ -5,6 +5,10 @@
 
 import * as crypto from 'crypto';
 
+// tslint:disable-next-line:variable-name
+const ZeroGuid = '00000000-0000-0000-0000-000000000000';
+export { ZeroGuid };
```
eanders-MS: prefix with export and make it a one liner.

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: Oh right, I forgot about that file.

packages/sdk/src/utils/deterministicGuids.ts
eanders-MS: This can be deleted now correct?

---
In reply to: [253243773](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/164#discussion_r253243773) [](ancestors = 253243773)

eanders-MS: Oh I see you backed it out.

---
In reply to: [253566140](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/164#discussion_r253566140) [](ancestors = 253566140,253243773)

packages/sdk/src/math/vector2.ts
eanders-MS: Oh it already had one. Missed that.

---
In reply to: [253243411](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/164#discussion_r253243411) [](ancestors = 253243411)

packages/sdk/src/types/runtime/assets/material.ts
```diff
@@ -50,6 +60,31 @@ export class Material extends Asset implements MaterialLike, Patchable<AssetLike
     /** @inheritdoc */
     public get color() { return this._color; }
```
eanders-MS: >public get color() { return this._color; } [](start = 4, length = 42)

Seems like you'd want to be able to set color too. In the setter it would just copy to the existing color instance, to preserve the observe hooks.

stevenvergenz: Oh, good idea!

PR #172, opened 2019-02-04
AssetManager.createMaterial, createTexture, ready

stevenvergenz:
* Make the asset constuctors take partials, and safely handle missing values.
* Add a `uri` field to `TextureLike`s for tracking texture sources separately from asset group sources.
* Refactor `AssetManager.ready` to be a promise chain instead of a dynamically created Promise.all call. New loads simply append their load to the chain. This dramatically reduces the complexity of the system.
* Added a default "manual" asset group that contains all individually created assets.
* Add `AssetManager.createMaterial(name: string, definition: MaterialLike): ForwardPromise<Material>`
* Add `AssetManager.createTexture(name: string, definition: TextureLike): ForwardPromise<Texture>`

### DIFF
--- a/packages/sdk/src/types/runtime/assets/texture.ts
+++ b/packages/sdk/src/types/runtime/assets/texture.ts
@@ -64,9 +64,18 @@ export class Texture extends Asset implements TextureLike, Patchable<AssetLike> 
             throw new Error("Cannot construct texture from non-texture definition");
         }
 
-        this._resolution = new Vector2(def.texture.resolution.x, def.texture.resolution.y);
-        this._wrapU = def.texture.wrapU;
-        this._wrapV = def.texture.wrapV;
+        if (def.texture.uri) {
+            this._uri = def.texture.uri;
+        }
+        if (def.texture.resolution) {
+            this._resolution = new Vector2(def.texture.resolution.x, def.texture.resolution.y);
+        }
+        if (def.texture.wrapU) {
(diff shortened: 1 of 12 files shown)


eanders-MS:
> export class Asset implements AssetLike {

unrelated but, make abstract class?

---
Refers to: packages/sdk/src/types/runtime/assets/asset.ts:55 in 1f3d4a2. [](commit_id = 1f3d4a2667fcb2cf2990fe70b49a73281f808646, deletion_comment = False)

packages/sdk/src/types/runtime/assets/texture.ts
```diff
@@ -28,6 +29,7 @@ export enum TextureWrapMode {
 
 export class Texture extends Asset implements TextureLike, Patchable<AssetLike> {
     // tslint:disable:variable-name
+    private _uri: string;
```
eanders-MS: >_uri [](start = 12, length = 4)

If this value is observed, it needs a default value to be assigned before the observe hooks are set. Otherwise, it won't be hooked (because it doesn't exist on the object until assigned).

stevenvergenz: It's not observed. In fact it doesn't have a setter app-side. It's assumed that this value is only set host-side.

stevenvergenz:
Paired Unity PR: https://github.com/Microsoft/mixed-reality-extension-unity/pull/22

Issue #99, opened 2019-01-07
Add 'light-test' functional test


eanders-MS:
We need a test for `enableLight` functionality.

PR #173, opened 2019-02-04
Add light-test, and related changes

eanders-MS:
The light component is one of the oldest features of our API but hasn't been used much, so it was interesting to revisit it. I found a few places where it could be improved, and these changes also applied to the other actor components: text and rigid body. In summary: I removed the `enable-*` payloads and instead relied on the actor patching mechanism to enable components. This was already 99% supported in the SDK, and 100% in the client, so the `enable-*` messages were unnecessary. This is a nice reduction in complexity! Checkout all the deleted code!

Fixes issue #99 

Clip of the light test in action
light-test

#### Other changes:
* The most recent tests I added follow a design pattern that allows the user to click on anything in the test to quit, or wait one minute. I updated my two previous tests to the latest version of this pattern.
* Added a static method to Quaternion to create a "look at" rotation.
* Updated the call signature of the `observe` method to be more self documenting.


packages/functional-tests/src/tests/reparent-test.ts
```diff
@@ -18,17 +18,20 @@ export default class ReparentTest extends Test {
...
+            this.sceneRoot = MRESDK.Actor.CreateEmpty(this.app.context).value;
+            const runningTestPromise = this.runTest();
+            const timeout = setTimeout(() => this.running = false, 60000);
+            await runningTestPromise;
+            clearTimeout(timeout);
+            return true;
+        } finally {
+            destroyActors(this.sceneRoot);
+        }
     }
```
sorenhan: managing actors shouldn't be needed anymore, as functional test menu wrapper cleans up all actors on exit. It's fine to do it manually, but I think we agreed unnecessary, right?

sorenhan: (just noting here because you added extra code around the destroy call)

---
In reply to: [254025155](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/173#discussion_r254025155) [](ancestors = 254025155)

eanders-MS: Oh that's right. I'll take this out.

---
In reply to: [254025313](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/173#discussion_r254025313) [](ancestors = 254025313,254025155)

packages/functional-tests/src/tests/reparent-test.ts
```diff
@@ -45,7 +48,9 @@ export default class ReparentTest extends Test {
                     color: MRESDK.Color3.Yellow()
                 }
             }
-        });
+        }).value;
+        label.setBehavior(MRESDK.ButtonBehavior).onClick('released', () => this.running = false);
+
```
sorenhan: is this label clickable wihtout having a collidable flag set anywhere? are colliders created by setbehavior, or automatically, or am I not reading this right?

eanders-MS: This is here for the future when colliders on empty objects with text are possible. Of course by that time, we'll probably have a generalized way to exit a test and it won't be necessary.

---
In reply to: [254026146](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/173#discussion_r254026146) [](ancestors = 254026146)

PR #179, opened 2019-02-06
Soren/update deploy notes

sorenhan:


packages/sdk/src/adapters/multipeer/adapter.ts
```diff
@@ -89,20 +89,20 @@ export class MultipeerAdapter extends Adapter {
                 sessionId,
                 connection: pipe.remote
             });
-            // Start the connection update loop (todo move this)
-            context.internal.start();
+            // Startup the context.
+            context.internal.start().catch(() => pipe.remote.close());
```
eanders-MS: This isn't right. There's a fix made here that I don't see.

packages/sdk/src/types/internal/context.ts
```diff
@@ -622,18 +584,58 @@ export class InternalContext {
...
+                execution.on('protocol.perform-action', this.performAction);
+                execution.on('protocol.receive-rpc', this.receiveRPC);
+                execution.on('protocol.collision-event-raised', this.collisionEventRaised);
+                execution.on('protocol.set-animation-state', this.setAnimationStateEventRaised);
+
+                // Startup the execution protocol
+                execution.startListening();
+
+                this.interval = setInterval(() => this.update(), 0);
+                this.context.emitter.emit('started');
```
eanders-MS: I split this out into its own method, but that change isn't here.

sorenhan:
I had the wrong base branch - ignore all comments on non-.md files

PR #181, opened 2019-02-07
Return existing asset group if load source is the same

stevenvergenz:
Fixes functional test errors on successive runs

### DIFF
--- a/packages/sdk/src/types/runtime/assets/assetManager.ts
+++ b/packages/sdk/src/types/runtime/assets/assetManager.ts
@@ -111,10 +111,9 @@ export class AssetManager {
         let group: AssetGroup;
         if (this.groups[groupName]) {
             group = this.groups[groupName];
-            if(group.source.containerType === 'gltf' && group.source.uri == uri) {
+            if (group.source.containerType === 'gltf' && group.source.uri === uri) {
                 return group;
-            }
-            else {
+            } else {
                 throw new Error(
                     `Group name ${groupName} is already in use. Unload the old group, or choose a different name.`);
             }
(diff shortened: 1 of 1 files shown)


packages/sdk/src/types/runtime/assets/assetManager.ts
```diff
@@ -108,12 +108,19 @@ export class AssetManager {
     private async loadGltfHelper(
         groupName: string, uri: string, colliderType: CreateColliderType): Promise<AssetGroup> {
 
+        let group: AssetGroup;
         if (this.groups[groupName]) {
-            throw new Error(
-                `Group name ${groupName} is already in use. Unload the old group, or choose a different name.`);
+            group = this.groups[groupName];
+            if(group.source.containerType === 'gltf' && group.source.uri == uri) {
```
eanders-MS: >== [](start = 73, length = 2)

Should be triple-equal: ===

eanders-MS: >if [](start = 12, length = 2)

Add a space after `if`

packages/sdk/src/types/runtime/assets/assetManager.ts
```diff
@@ -108,12 +108,19 @@ export class AssetManager {
...
 
+        let group: AssetGroup;
         if (this.groups[groupName]) {
-            throw new Error(
-                `Group name ${groupName} is already in use. Unload the old group, or choose a different name.`);
+            group = this.groups[groupName];
+            if(group.source.containerType === 'gltf' && group.source.uri == uri) {
+                return group;
+            }
+            else {
```
eanders-MS: Close bracket and `else` on same line. I recommend running the format document command in VSCode.

PR #182, opened 2019-02-07
Asset preload test - Only clean up actors successfully created

stevenvergenz:


packages/functional-tests/src/tests/gltf-concurrency-test.ts
```diff
@@ -57,9 +61,10 @@ export default class GltfConcurrencyTest extends Test {
         const bottle = await MRESDK.Actor.CreateFromPrefab(this.app.context, {
             prefabId: bottleAsset.prefabs.byIndex(0).id
         });
+        actors.push(bottle);
 
         await delay(10000);
-        destroyActors([runner, gearbox, bottle]);
+        destroyActors(actors);
```
eanders-MS: The app handles actor destruction now. You shouldn't have to do this.

stevenvergenz: So I can remove all the cleanup logic?

Issue #188, opened 2019-02-12
Multi-peer bug: Queued actions can be sent before actor is ready.


eanders-MS:
The SDK has an internal mechanism for waiting until an actor has been created on the host before allowing other queued actions through, like creating animations, adding a rigid body, etc. Problem is it only waits for it to be created on one peer, not all of them. This means subsequent calls can fail on some peers, depending whether or not the actor got created in time.

PR #189, opened 2019-02-13
[SDK] Fixes for stability

eanders-MS:
Found and fixed numerous issues exposed by stressing network traffic in different ways.
- Fix for #188: "Multi-peer bug: Queued actions can be sent before actor is ready." When the app sends a message that expects a reply, the multipeer adapter will now wait for replies from all peers before forwarding the reply back to the app.
- There was a small window of time between client 'handshake' and 'startup' steps where a message could get dropped. It's a weakness in the protocol design. Added a workaround for now. Will revisit the design later.
- During animation synchronization there's an edge case that could result in an unexpected reply message. Added a workaround for that. This issue will go away soon with the planned anim sync refactor.
- All messages that expect a reply must be able to handle a response type of `operation-result`. Updated this everywhere.
- `createAnimation` was expecting the wrong response message type.
- `setAnimationState` now waits until the actor is created. 

Also:
- Changed network logging so that logging to the 'network' facility shows a summary of network traffic. To see the contents of network traffic, use `log.enable('network-content');`

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,145 +1,36 @@
 {
-  // See https://go.microsoft.com/fwlink/?LinkId=733558
-  // for the documentation about the tasks.json format
-  "version": "2.0.0",
-  "tasks": [
-    {
-      "type": "npm",
-      "script": "build",
-      "path": "packages/",
-      "presentation": {
-        "echo": true,
-        "reveal": "always",
-        "focus": false,
-        "panel": "shared",
-        "showReuseMessage": true,
-        "clear": true
-      },
-      "problemMatcher": []
-    },
-    {
-      "type": "npm",
-      "script": "build",
-      "path": "packages/sdk/",
-      "presentation": {
-        "echo": true,
-        "reveal": "always",
-        "focus": false,
-        "panel": "shared",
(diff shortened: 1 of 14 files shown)


eanders-MS:
>                         if (session.isAnimating(syncActor)) {

Mystery here. We want to let the transform propagate if a rigid body exists on the Actor. That seems to work correctly in local tests, but this code suggests it shouldn't. Will figure this out and update the PR(s).

---
Refers to: packages/sdk/src/adapters/multipeer/rules.ts:298 in 92188c4. [](commit_id = 92188c476fdb8e74b37156df8f355d6a4982d10e, deletion_comment = False)

eanders-MS:
>                         if (session.isAnimating(syncActor)) {

This is limited to rigid bodies that are animating... an edge case so not worried now.

---
In reply to: [463092543](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/189#issuecomment-463092543) [](ancestors = 463092543)

---
Refers to: packages/sdk/src/adapters/multipeer/rules.ts:298 in 92188c4. [](commit_id = 92188c476fdb8e74b37156df8f355d6a4982d10e, deletion_comment = False)

PR #191, opened 2019-02-13
[NOT FOR MERGE] Proposed Sound API

sorenhan:
Here's what I think the Sound API could look like.

Please read and comment, especially on the sound-design.MD file.

### DIFF
--- a/packages/functional-tests/src/tests/sound-test.ts
+++ b/packages/functional-tests/src/tests/sound-test.ts
@@ -1,0 +1,102 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import * as MRESDK from '@microsoft/mixed-reality-extension-sdk';
+import { Actor } from '@microsoft/mixed-reality-extension-sdk';
+import App from '../app';
+import delay from '../utils/delay';
+import destroyActors from '../utils/destroyActors';
+import Test from './test';
+
+export default class SoundTest extends Test {
+
+    constructor(app: App, private baseUrl: string) {
+        super(app);
+    }
+
+    public async run(): Promise<boolean> {
+        let success = true;
+
+        success = success && await this.runSoundTest();
+
+        return success;
+    }
+
+    public async runSoundTest(): Promise<boolean> {
+        // Make a root object.
+        const tester = MRESDK.Actor.CreateEmpty(this.app.context, {});
+
+        const textPromise = Actor.CreateEmpty(this.app.context, {
+            actor: {
+                name: 'label',
+                parentId: tester.value.id,
+                transform: {
+                    position: { x: 0, y: 2, z: 0 }
+                },
+                text: {
+                    anchor: MRESDK.TextAnchorLocation.MiddleCenter,
+                    color: { r: 30 / 255, g: 206 / 255, b: 213 / 255 },
+                    height: 0.3
+                }
+            }
+        });
+
+        const text = textPromise.value;
+        await textPromise;
+        text.text.contents = "Preloading Sound";
+
+        const soundId = this.app.context.assetManager.loadSound(
+            'sounds', `${this.baseUrl}/testsound.wav`,
+            {
+                looping: true
+            });
+
+        await soundId;
+        await delay(1 * 1000);
+
+        text.text.contents = "Starting Sound";
+        const instanceHandle = text.playSound(await soundId,
+            {
+                volume: 0.5
+            },
+            42);
+        await delay(3 * 1000);
+
+        text.text.contents = "Increasing Pitch by 12 semitones";
+        MRESDK.Sound.setSoundState(instanceHandle, {
+            pitch: 12.0,
+        });
+        await delay(3 * 1000);
+
+        text.text.contents = "Raising Volume";
+        MRESDK.Sound.setSoundState(instanceHandle, {
+            volume: 1.0
+        });
+        await delay(3 * 1000);
+
+        text.text.contents = "Pausing";
+        MRESDK.Sound.pause(instanceHandle);
+        await delay(3 * 1000);
+
+        text.text.contents = "Playing non-preloaded sound";
+        const sound2 = text.loadAndPlaySound(`${this.baseUrl}/testsound2.wav`,
+            {},
+            {
+                volume: 0.5
+            },
+            41);
+        await delay(3 * 1000);
+
+        text.text.contents = "resuming";
+        MRESDK.Sound.resume(instanceHandle);
+        await delay(3 * 1000);
+
+        text.text.contents = "Stopping Sound";
+        MRESDK.Sound.stop(instanceHandle);
+        await delay(6 * 1000);
+
+        return true;
+    }
+}
--- a/packages/functional-tests/src/tests/sound-test.ts
+++ b/packages/functional-tests/src/tests/sound-test.ts
@@ -47,17 +47,17 @@ export default class SoundTest extends Test {
         await textPromise;
         text.text.contents = "Preloading Sound";
 
-        const soundId = this.app.context.assetManager.loadSound(
+        const soundAssetPromise = this.app.context.assetManager.loadSoundAsset(
             'sounds', `${this.baseUrl}/testsound.wav`,
             {
                 looping: true
             });
 
-        await soundId;
-        await delay(1 * 1000);
+        // await soundAssetPromise;
+        // await delay(1 * 1000);
 
         text.text.contents = "Starting Sound";
-        const instanceHandle = text.playSound(await soundId,
+        const soundInstance = text.playSound(soundAssetPromise.value,
             {
                 volume: 0.5
             },
@@ -65,36 +65,39 @@ export default class SoundTest extends Test {
         await delay(3 * 1000);
 
         text.text.contents = "Increasing Pitch by 12 semitones";
-        MRESDK.Sound.setSoundState(instanceHandle, {
+        soundInstance.value.setSoundState({
             pitch: 12.0,
         });
         await delay(3 * 1000);
 
         text.text.contents = "Raising Volume";
-        MRESDK.Sound.setSoundState(instanceHandle, {
+        soundInstance.value.setSoundState({
             volume: 1.0
         });
         await delay(3 * 1000);
 
         text.text.contents = "Pausing";
-        MRESDK.Sound.pause(instanceHandle);
+        soundInstance.value.pause();
         await delay(3 * 1000);
 
         text.text.contents = "Playing non-preloaded sound";
-        const sound2 = text.loadAndPlaySound(`${this.baseUrl}/testsound2.wav`,
+        const soundInstance2 = text.loadAndPlaySound(`${this.baseUrl}/testsound2.wav`,
             {},
             {
                 volume: 0.5
             },
             41);
         await delay(3 * 1000);
-
+        soundInstance2.value.stop();
+        await delay(1 * 1000);
         text.text.contents = "resuming";
-        MRESDK.Sound.resume(instanceHandle);
+        soundInstance.value.resume();
         await delay(3 * 1000);
 
         text.text.contents = "Stopping Sound";
-        MRESDK.Sound.stop(instanceHandle);
+        soundInstance.value.stop();
+        this.app.context.assetManager.unloadSoundAsset(soundAssetPromise.value);
+
         await delay(6 * 1000);
 
         return true;
--- a/packages/sdk/src/sound/setSoundStateOptions.ts
+++ b/packages/sdk/src/sound/setSoundStateOptions.ts
@@ -18,12 +18,8 @@ export type SetSoundStateOptions = {
     pitch?: number,
 
     /**
-     * volume modifier, measured in DB (0=default,10=10x power, -10=1/10 power.
+     * volume multiplier, (0.0-1.0, where 0.0=no sound, 1.0=maximum)
      */
     volume?: number,
 
-    /**
-     * time in milliseconds
-     */
-    time?: number,
 };
--- a/packages/sdk/src/sound/setSoundStateOptions.ts
+++ b/packages/sdk/src/sound/setSoundStateOptions.ts
@@ -1,0 +1,29 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Parameters to the `Sound.setSoundState` and `Actor.playSound` calls.
+ */
+export type SetSoundStateOptions = {
+    /**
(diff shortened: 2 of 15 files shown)


sound-design.MD
```diff
@@ -0,0 +1,42 @@
+# MRE Sound API proposal
+
+This is node.js side only. It doesn't actually do anything yet.
+
+## Required functionality
+* preload sounds (as looping or one-shot)
+* play preloaded sounds
+* load and play sounds (Basically preload, instantly play, unload upon completion)
```
stevenvergenz: >load and play sounds [](start = 2, length = 20)

I wonder if the latency of loading the sound would nullify the utility of this method. Don't most sounds require timing?

sound-design.MD
```diff
@@ -0,0 +1,42 @@
...
+
+## Design thoughts
+
+(Feel free to comment on each item separately)
+
+I'm using assetmanager to for the preloading sound to ask the client to load a sound resource, and return a handle. This is therefore are not instant, but return a promise.
+
+All sounds play on an actor, and the play calls therefore happen on the actor.
+
+PlaySound are played by passing in a sound resource ID, and instantly return a sound instance ID.
```
stevenvergenz: >instantly return [](start = 60, length = 16)

Via a ForwardPromise, like Actors?

sound-design.MD
```diff
@@ -0,0 +1,42 @@
...
+
+LoadAndPlaySound takes a URI, and returns a promise to a sound instance ID when the primary client has started playing the sound.
+
+The sound instance doesn't store any data on the server at all, it lets the client manage the instance fully. It's up to the client to take the sound instance handle and tweak it if necessary. Sound instance handles expire the moment a non-looping sound reaches the end, or (for both looping/non-looping sounds) when the user calls stop() for that sound instance.
```
stevenvergenz: >ets the client manage the instance fully [](start = 68, length = 40)

This would make sounds heavily dependent on the authoritative peer. Seems like it would make sense to at least save clip duration and start time server-side, so we can estimate completion time and auto-expire instances without an extra round trip.

sound-design.MD
```diff
@@ -0,0 +1,42 @@
...
+* stop sounds 
+* set the sound state (to change volume, pitch, and where possible seek)
+
+## Design thoughts
+
+(Feel free to comment on each item separately)
+
+I'm using assetmanager to for the preloading sound to ask the client to load a sound resource, and return a handle. This is therefore are not instant, but return a promise.
+
+All sounds play on an actor, and the play calls therefore happen on the actor.
```
stevenvergenz: >play calls therefore happen on the actor [](start = 37, length = 40)

Is there a limit to the number of sounds that can live on a single actor? Might get confusing routing commands to a particular instance in this case.

sound-design.MD
```diff
@@ -0,0 +1,42 @@
...
+We may want to add "3d-ness" to the sound state, too, but that's not super important now.
+
+
+## I'd especially like feedback on:
+
+This API has been designed to follow conventions of other parts of the SDK (AssetManager for preloading,  Actor/Animation for starting/stopping/changing state). I'm very happy to change it where recommended
+
+Should I switch over to allocating a data structure on the node side for managing the state? I don't think that'll buy anything, which is why I went with the "here's a handle".
+
+I guess I need an unload preloaded sound command, right?
```
stevenvergenz: >unload preloaded sound [](start = 18, length = 22)

We need an unload preloaded *anything* command right now.

sound-design.MD
```diff
@@ -0,0 +1,42 @@
...
+
+## I'd especially like feedback on:
+
+This API has been designed to follow conventions of other parts of the SDK (AssetManager for preloading,  Actor/Animation for starting/stopping/changing state). I'm very happy to change it where recommended
+
+Should I switch over to allocating a data structure on the node side for managing the state? I don't think that'll buy anything, which is why I went with the "here's a handle".
+
+I guess I need an unload preloaded sound command, right? 
+
+What's best terminology? SoundResource/SoundInstance? Unity uses "AudioSource" and "AudioClip", and they seem non-intuitive to me, as a source
```
stevenvergenz: >SoundResource/SoundInstance [](start = 25, length = 27)

Unity => SDK, AudioClip => SoundAsset, AudioSource => SoundInstance

packages/functional-tests/src/tests/sound-test.ts
```diff
@@ -0,0 +1,102 @@
...
+                    height: 0.3
+                }
+            }
+        });
+
+        const text = textPromise.value;
+        await textPromise;
+        text.text.contents = "Preloading Sound";
+
+        const soundId = this.app.context.assetManager.loadSound(
```
stevenvergenz: Currently we use the "load" term to refer to calls that return AssetGroups, and "create" to refer to calls that return Assets. Not sure if we want to keep this, but that's what in place currently.

sound-design.MD
sorenhan: SoundAsset, I like it!.

---
In reply to: [256946579](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256946579) [](ancestors = 256946579)

packages/functional-tests/src/tests/sound-test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+        await soundId;
+        await delay(1 * 1000);
+
+        text.text.contents = "Starting Sound";
+        const instanceHandle = text.playSound(await soundId,
+            {
+                volume: 0.5
+            },
+            42);
```
stevenvergenz: What is "priority"?

packages/functional-tests/src/tests/sound-test.ts
```diff
@@ -0,0 +1,102 @@
...
+        await delay(3 * 1000);
+
+        text.text.contents = "Increasing Pitch by 12 semitones";
+        MRESDK.Sound.setSoundState(instanceHandle, {
```
stevenvergenz: >MRESDK.Sound.setSoundState(instanceHandle [](start = 8, length = 41)

Seems reasonable to make this an instance method, and then have it call into AssetManager maybe. That's how material/texture state is done.

packages/sdk/src/sound/setSoundStateOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+     */
+    enabled?: boolean,
+
+    /**
+     * pitch offset in halftones (0=default, 12=one octave higher, -12=one octave lower)
+     */
+    pitch?: number,
+
+    /**
+     * volume modifier, measured in DB (0=default,10=10x power, -10=1/10 power.
```
stevenvergenz: >volume modifier, measured in DB [](start = 7, length = 31)

This isn't what I expected. Would prefer a 0-1 volume percentage, seems more intuitive.

packages/functional-tests/src/tests/sound-test.ts
```diff
@@ -0,0 +1,102 @@
...
+        await textPromise;
+        text.text.contents = "Preloading Sound";
+
+        const soundId = this.app.context.assetManager.loadSound(
+            'sounds', `${this.baseUrl}/testsound.wav`,
+            {
+                looping: true
+            });
+
+        await soundId;
```
eanders-MS: >await soundId; [](start = 8, length = 14)

Can you rewrite this test to not require awaiting soundId? text.playSound should be immediately callable. (actor should await on the sound to load before playing it). This is consistent with the way animations work.

packages/sdk/src/sound/setSoundStateOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+     */
+    volume?: number,
+
+    /**
+     * time in milliseconds
```
stevenvergenz: >time in milliseconds [](start = 7, length = 20)

Does this mean the app needs to know the length of the clip in order to load it?

packages/functional-tests/src/tests/sound-test.ts
eanders-MS: >loadSound [](start = 54, length = 9)

loadSound should return a ForwardPromise<SoundAsset>, so the app can immediately act on it. This would be consistent with how Actor and Animation work.

packages/sdk/src/sound/setSoundStateOptions.ts
eanders-MS: >milliseconds [](start = 15, length = 12)

In the animation system, time is expressed in seconds. I'd like to be consistent across the API and standardize on seconds unless there's a compelling reason for milliseconds here.

sorenhan: good point

---
In reply to: [256951167](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256951167) [](ancestors = 256951167)

packages/functional-tests/src/tests/sound-test.ts
```diff
@@ -0,0 +1,102 @@
...
+        await delay(1 * 1000);
+
+        text.text.contents = "Starting Sound";
+        const instanceHandle = text.playSound(await soundId,
```
eanders-MS: >instanceHandle [](start = 14, length = 14)

I think an interface wrapper would be a more appropriate design here. Something like `SoundInstance`.
```ts
class SoundInstance {
    public setSoundState(state: Partial<SoundState>);
    public play();
    public pause();
    ...etc
};
```

eanders-MS: I'm thinking the same may apply to animations. I don't love that the animation API is on Actor.

---
In reply to: [256952297](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256952297) [](ancestors = 256952297)

packages/functional-tests/src/tests/sound-test.ts
sorenhan: That seems strange to me. "load" to me means getting from an external location, while "create" indicates generating from scratch. I guess that "create" to return asset may make sense if it's creating a handle for the asset out of something already loaded, but that's not the case here.

sound-design.MD
sorenhan: doesn't seem like there is a limit. I don't think it's confusing as they are accessed from the instance, not from the actor.

sound-design.MD
sorenhan: I think this may be a useful utility sound for some simple things, but not the method that gives the most control.It's the equivalent of CreateFromGLTF vs instantiating preloaded GLTF, I think it's useful.

packages/sdk/src/sound/setSoundStateOptions.ts
sorenhan: yeah, I was getting too creative here. Will restore to 0-1

---
In reply to: [256950001](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256950001) [](ancestors = 256950001)

packages/functional-tests/src/tests/sound-test.ts
sorenhan: Useful if we have too many sounds and have to cull out some. But true, it's not important now, so cutting it

---
In reply to: [256948754](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256948754) [](ancestors = 256948754)

packages/sdk/src/sound/setSoundStateOptions.ts
sorenhan: Resolved. I moved time to be on PlaySound/LoadAndPlaySound, rather than here,. but changed to be seconds

---
In reply to: [256951973](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256951973) [](ancestors = 256951973,256951167)

packages/functional-tests/src/tests/sound-test.ts
sorenhan: I agree this is better for soundinstance. Don't know if it's better for animations - unlike sound, there's not really animation instance handles created/destroyed at start/stop. And animation names are not synced from client to server, right? I guess there are workarounds for that, though. Regardless, this PR is for sound, so and I'll follow this pattern.

---
In reply to: [256953209](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256953209) [](ancestors = 256953209,256952297)

packages/functional-tests/src/tests/sound-test.ts
tombuMS: Isn't that what "instantiate" calls were introduced to solve?

sorenhan: I'm clarifying now that this is the soundasset

---
In reply to: [256979671](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256979671) [](ancestors = 256979671)

sound-design.MD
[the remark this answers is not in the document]
sorenhan: I was debating this. I think that until we have a way to give permission for global changes (like directional light, ambient, skybox), we wouldn't support global sounds. In the AltspaceVR case, it would be fine to support this, but this doesn't truly fit one of the main ideas behind MRE, that MREs are holograms of some finite size

---
In reply to: [256634664](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256634664) [](ancestors = 256634664)

packages/functional-tests/src/tests/sound-test.ts
sorenhan: yeah, i'm changing this. thanks

---
In reply to: [256949147](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256949147) [](ancestors = 256949147)

sound-design.MD
sorenhan: that's not originally what I was thinking, but I agree that forwardpromise is the right thing

---
In reply to: [256944035](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256944035) [](ancestors = 256944035)

sound-design.MD
eanders-MS: I don't think the server needs to know the length of the clip. It just needs to know when it started playing and when has stopped. The authoritative client would need to send a message to let the server know the sound has ended.

---
In reply to: [256945051](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r256945051) [](ancestors = 256945051)

sorenhan: Yeah, I'm changing the design here to have it managed by the server instead. And yeah, it will be relying on stop messages

---
In reply to: [257022732](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/191#discussion_r257022732) [](ancestors = 257022732,256945051)

sorenhan:
Thank you for all the feedback. Closing this as implementation is underway

PR #192, opened 2019-02-14
[SDK] More networking fixes

eanders-MS:
This PR contains three bugfixes:
1. A race condition in the client join process vs. how we were tracking the authoritative peer could cause a mismatch between which peer the server considered authoritative and which client was acting as the authoritative peer. Bug symptoms include: Behaviors not responding on non-authoritative instances. We've seen this a lot in the wild, where something isn't clickable. Really great to have tracked this one down.
2. A regression introduced in my last PR would cause every non-authoritative peer to repeatedly disconnect after a minute then reconnect. Doh!
3. A regression introduced in my last PR where client replies weren't getting preprocessed, resulting in gaps in the cached synchronization state. Doh!

To whoever added the Connect and Disconnect buttons to the MREComponent in the MRETestBed, THANK YOU! Super helpful.

### DIFF
--- a/packages/sdk/src/adapters/multipeer/session.ts
+++ b/packages/sdk/src/adapters/multipeer/session.ts
@@ -28,7 +28,9 @@ export class Session extends EventEmitter {
     public get conn() { return this._conn; }
     public get sessionId() { return this._sessionId; }
     public get protocol() { return this._protocol; }
-    public get clients() { return Object.keys(this._clientSet).map(clientId => this._clientSet[clientId]); }
+    public get clients() {
+        return Object.keys(this._clientSet).map(clientId =>
+            this._clientSet[clientId]).sort((a, b) => a.order - b.order); }
     public get actors() { return Object.keys(this._actorSet).map(actorId => this._actorSet[actorId]); }
     public get assets() { return this._assets; }
     public get assetUpdates() {
@@ -39,7 +41,7 @@ export class Session extends EventEmitter {
             !this._actorSet[actorId].created.message.payload.actor.parentId).map(actorId => this._actorSet[actorId]);
     }
     public get users() { return Object.keys(this._userSet).map(userId => this._userSet[userId]); }
-    public get authoritativeClient() { return this.clients.sort((a, b) => a.order - b.order).shift(); }
+    public get authoritativeClient() { return this.clients.find(client => client.authoritative); }
     public get peerAuthoritative() { return this._peerAuthoritative; }
     public get actorSet() { return this._actorSet; }
     public get assetUpdateSet() { return this._assetUpdateSet; }
@@ -114,11 +116,9 @@ export class Session extends EventEmitter {
             // Once the client is joined, further messages from the client will be processed by the session
             // (as opposed to a protocol class).
             client.on('recv', (_: Client, message: Message) => this.recvFromClient(client, message));
-            if (this.authoritativeClient) {
-                this.authoritativeClient.sendPayload({
-                    type: 'set-authoritative',
-                    authoritative: this.peerAuthoritative,
-                } as Payloads.SetAuthoritative);
+            // If we don't have an authoritative client, make this client authoritative.
+            if (!this.authoritativeClient) {
+                this.setAuthoritativeClient(client.id);
             }
         } catch (e) {
             log.error('network', e);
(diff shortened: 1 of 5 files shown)


packages/sdk/src/adapters/multipeer/session.ts
```diff
@@ -206,14 +218,18 @@ export class Session extends EventEmitter {
...
                     client.send({ ...message }, {
                         resolve: (payload: Payloads.Payload, response: Message) => {
-                            // Cache the response message.
-                            if (client.authoritative) {
-                                authoritativeReplyMessage = {...response };
-                            } else if (!fallbackReplyMessage) {
-                                fallbackReplyMessage = { ...response };
-                            }
                             // Decrement the wait counter.
                             awaitCount -= 1;
```
stevenvergenz: >awaitCount -= 1; [](start = 28, length = 16)

What happens when a client disconnects during this process? It might be better to see if the number of responses is equal to the number of connected clients or something.

eanders-MS: When a client disconnects the reject handler (below) is called and the awaitCount decremented there.

---
In reply to: [256942555](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/192#discussion_r256942555) [](ancestors = 256942555)

sorenhan: If that was changed then there could the opposite problem - if a disconnecting client has already received the response. 
Does disconnecting trigger a response? if so, then Eric's code would always reach 0

---
In reply to: [256942555](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/192#discussion_r256942555) [](ancestors = 256942555)

sorenhan: Eric answering my question before I even asked it :-)

---
In reply to: [256943726](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/192#discussion_r256943726) [](ancestors = 256943726,256942555)

eanders-MS: It's a tricky piece of code for sure.

---
In reply to: [256943945](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/192#discussion_r256943945) [](ancestors = 256943945,256943726,256942555)

packages/sdk/src/adapters/multipeer/session.ts
```diff
@@ -156,6 +157,17 @@ export class Session extends EventEmitter {
         } catch { }
     }
 
+    private setAuthoritativeClient(clientId: string) {
+        if (!this._clientSet[clientId]) {
```
sorenhan: Not required, but I think some MREs could benefit from having this public. If MREs can set the authoritative client, and if physics simulation is always better on the primary machine, things with physics and turn-based interactions (like bowling, darts, golf/mini-golf,  skeeball, rolling dice etc), would look great and feel more responsive for the active user

eanders-MS: I hesitate to make MREs aware of the underlying operating model of the network. It has no idea there are multiple connections or multiple peers currently.

Each connection maintains an exponential moving average of its latency. If we also knew the connection's platform, the multipeer adapter could make a pretty informed choice for who should be the authoritative peer.

---
In reply to: [256946008](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/192#discussion_r256946008) [](ancestors = 256946008)

sorenhan: Yeah, I agree with that goal. Maybe it could be more of a hint that the underlying system can choose to apply if it wants. (SetPrimaryInteractingUser)?

---
In reply to: [256949161](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/192#discussion_r256949161) [](ancestors = 256949161,256946008)

sorenhan: (But this is not really related to this PR, so closing this.)

---
In reply to: [256950392](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/192#discussion_r256950392) [](ancestors = 256950392,256949161,256946008)

eanders-MS:
@stevenvergenz Are you requesting changes?

PR #195, opened 2019-02-15
[SDK] Remove unused code

eanders-MS:
Removed the notion of user subscriptions and user transform. We weren't using them, and probably never will.


stevenvergenz:
You don't think we'll ever need to know where a user is? Sounds useful to me. All these relate to that state update.

eanders-MS:
We spoke offline... the attachment system offers a way to get the user's position (by attaching an empty object to an attach point). The main thing is to not have multiple ways for this info to be synced over the network. If the attach point method trips up users, we can add a convenience layer.

---
In reply to: [464143600](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/195#issuecomment-464143600) [](ancestors = 464143600)

PR #203, opened 2019-02-16
Update sdk to update actor to allow enabling collider on an actor.

tombuMS:
Updated the API for enabling colliders to use the actor patching mechanism instead of the separate payload for enabling a collider.

### DIFF
--- a/packages/sdk/src/types/runtime/actor.ts
+++ b/packages/sdk/src/types/runtime/actor.ts
@@ -6,6 +6,7 @@
 import events from 'events';
 import {
     Collider,
+    CollisionLayer,
     ColliderLike,
     CollisionData,
     Light,
@@ -24,7 +25,8 @@ import {
     CreateAnimationOptions,
     LookAtMode,
     PrimitiveDefinition,
-    SetAnimationStateOptions
+    SetAnimationStateOptions,
+    Vector3Like
 } from '../..';
 import { ZeroGuid } from '../../constants';
 import { log } from '../../log';
@@ -36,6 +38,7 @@ import { CollisionEventType, CreateColliderType } from '../network/payloads';
 import { SubscriptionType } from '../network/subscriptionType';
 import { Patchable } from '../patchable';
 import { Behavior } from './behaviors';
+import { BoxColliderParams, ColliderParams, SphereColliderParams } from './physics';
 
 /**
  * Describes the properties of an Actor.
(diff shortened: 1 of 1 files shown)


packages/sdk/src/types/runtime/actor.ts
```diff
@@ -311,59 +314,67 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+            const colliderParams = this._generateColliderParams(colliderType, center, size);
+
+            if (colliderParams) {
+                this._collider.copy({
+                    enabled: true,
+                    isTrigger,
+                    collisionLayer,
+                    colliderParams
+                } as ColliderLike)
+            }
```
eanders-MS: The other components unconditionally copy the values (regardless of whether the component existed already). Would be good to do that here to, for consistency. Is there a reason not to?

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -657,4 +668,32 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
                 .catch(reason => log.error('app', reason));
         }
     }
+
+    private _generateColliderParams(
```
eanders-MS: >_generateColliderParams [](start = 12, length = 23)

The leading underscore should cause a lint failure. It's fine to disable it for this line. I'm eventually going to update the tslint rule to allow underscore on private members.

packages/sdk/src/types/runtime/actor.ts
tombuMS: The collider params are immutable, so they shouldn't be set if there already exists a collider.  We had discussed this the other day.

packages/sdk/src/types/runtime/actor.ts
tombuMS: There is no linter failure that happened for me.  I though we were using this for private function names.  Either way, no error for me.

packages/sdk/src/types/runtime/actor.ts
eanders-MS: Ok. I didn't realize all of them were immutable.

---
In reply to: [257434402](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/203#discussion_r257434402) [](ancestors = 257434402)

packages/sdk/src/types/runtime/actor.ts
eanders-MS: OK. It is fine to use them on private functions if you want to. The lint rule needs some adjustment.

---
In reply to: [257434859](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/203#discussion_r257434859) [](ancestors = 257434859)

eanders-MS:
Is this something you plan to merge? Or just wanted to code review? I wish we could open code reviews on branches without having to make it a PR.

tombuMS:
> Is this something you plan to merge? Or just wanted to code review? I wish we could open code reviews on branches without having to make it a PR.

Yes.  This is a PR.  Is there something I should be adding to indicate this or something?  I did not mark it not merge.

eanders-MS:
👍

---
In reply to: [464264858](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/203#issuecomment-464264858) [](ancestors = 464264858)

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -311,59 +314,67 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+     */
+    public enableCollider(
+        colliderType: 'box',
+        collisionLayer: CollisionLayer,
+        isTrigger: boolean,
+        center?: Vector3Like,
+        size?: Vector3Like): void;
+
+    /** @ignore */
+    public enableCollider(
```
stevenvergenz: It could be useful to store the collider options somewhere in the actor for retrieval later. Not sure how frequently it'll be useful, but seems like good practice.

tombuMS: Not sure what options you are referring to.  The collider params?

stevenvergenz: Right, the collider params. Since collider settings are immutable, it's conceivable that a user might pool and reuse collider actors. Having after-the-fact access to the config would be useful for this case.

tombuMS:
Closing this PR in favor of a new one after some design changes to the API.

PR #204, opened 2019-02-16
[DO NOT MERGE] sound interim review

sorenhan:
Matches this Unity PR https://github.com/Microsoft/mixed-reality-extension-unity/pull/32

Please use sound-design.md for overall design comments

### DIFF
--- a/packages/functional-tests/src/tests/sound-test.ts
+++ b/packages/functional-tests/src/tests/sound-test.ts
@@ -1,0 +1,102 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import * as MRESDK from '@microsoft/mixed-reality-extension-sdk';
+import { Actor } from '@microsoft/mixed-reality-extension-sdk';
+import App from '../app';
+import delay from '../utils/delay';
+import destroyActors from '../utils/destroyActors';
+import Test from './test';
+
+export default class SoundTest extends Test {
+
+    constructor(app: App, private baseUrl: string) {
+        super(app);
+    }
+
+    public async run(): Promise<boolean> {
+        let success = true;
+
+        success = success && await this.runSoundTest();
+
+        return success;
+    }
+
+    public async runSoundTest(): Promise<boolean> {
+        // Make a root object.
+        const tester = MRESDK.Actor.CreateEmpty(this.app.context, {});
+
+        const textPromise = Actor.CreateEmpty(this.app.context, {
+            actor: {
+                name: 'label',
+                parentId: tester.value.id,
+                transform: {
+                    position: { x: 0, y: 2, z: 0 }
+                },
+                text: {
+                    anchor: MRESDK.TextAnchorLocation.MiddleCenter,
+                    color: { r: 30 / 255, g: 206 / 255, b: 213 / 255 },
+                    height: 0.3
+                }
+            }
+        });
+
+        const text = textPromise.value;
+        await textPromise;
+        text.text.contents = "Preloading Sound";
+
+        const soundId = this.app.context.assetManager.loadSound(
+            'sounds', `${this.baseUrl}/testsound.wav`,
+            {
+                looping: true
+            });
+
+        await soundId;
+        await delay(1 * 1000);
+
+        text.text.contents = "Starting Sound";
+        const instanceHandle = text.playSound(await soundId,
+            {
+                volume: 0.5
+            },
+            42);
+        await delay(3 * 1000);
+
+        text.text.contents = "Increasing Pitch by 12 semitones";
+        MRESDK.Sound.setSoundState(instanceHandle, {
+            pitch: 12.0,
+        });
+        await delay(3 * 1000);
+
+        text.text.contents = "Raising Volume";
+        MRESDK.Sound.setSoundState(instanceHandle, {
+            volume: 1.0
+        });
+        await delay(3 * 1000);
+
+        text.text.contents = "Pausing";
+        MRESDK.Sound.pause(instanceHandle);
+        await delay(3 * 1000);
+
+        text.text.contents = "Playing non-preloaded sound";
+        const sound2 = text.loadAndPlaySound(`${this.baseUrl}/testsound2.wav`,
+            {},
+            {
+                volume: 0.5
+            },
+            41);
+        await delay(3 * 1000);
+
+        text.text.contents = "resuming";
+        MRESDK.Sound.resume(instanceHandle);
+        await delay(3 * 1000);
+
+        text.text.contents = "Stopping Sound";
+        MRESDK.Sound.stop(instanceHandle);
+        await delay(6 * 1000);
+
+        return true;
+    }
+}
--- a/packages/functional-tests/src/tests/sound-test.ts
+++ b/packages/functional-tests/src/tests/sound-test.ts
@@ -48,7 +48,7 @@ export default class SoundTest extends Test {
         text.text.contents = "Preloading Sound";
 
         const soundAssetPromise = this.app.context.assetManager.loadSoundAsset(
-            'sounds', `${this.baseUrl}/testsound.wav`,
+            'sounds', `${this.baseUrl}/music.mod`,
             {
                 looping: true
             });
--- a/packages/functional-tests/src/tests/sound-test.ts
+++ b/packages/functional-tests/src/tests/sound-test.ts
@@ -47,17 +47,17 @@ export default class SoundTest extends Test {
         await textPromise;
         text.text.contents = "Preloading Sound";
 
-        const soundId = this.app.context.assetManager.loadSound(
+        const soundAssetPromise = this.app.context.assetManager.loadSoundAsset(
             'sounds', `${this.baseUrl}/testsound.wav`,
             {
                 looping: true
             });
 
-        await soundId;
-        await delay(1 * 1000);
+        // await soundAssetPromise;
+        // await delay(1 * 1000);
 
         text.text.contents = "Starting Sound";
-        const instanceHandle = text.playSound(await soundId,
+        const soundInstance = text.playSound(soundAssetPromise.value,
             {
                 volume: 0.5
             },
@@ -65,36 +65,39 @@ export default class SoundTest extends Test {
         await delay(3 * 1000);
 
         text.text.contents = "Increasing Pitch by 12 semitones";
-        MRESDK.Sound.setSoundState(instanceHandle, {
+        soundInstance.value.setSoundState({
             pitch: 12.0,
         });
         await delay(3 * 1000);
 
         text.text.contents = "Raising Volume";
-        MRESDK.Sound.setSoundState(instanceHandle, {
+        soundInstance.value.setSoundState({
             volume: 1.0
         });
         await delay(3 * 1000);
 
         text.text.contents = "Pausing";
-        MRESDK.Sound.pause(instanceHandle);
+        soundInstance.value.pause();
         await delay(3 * 1000);
 
         text.text.contents = "Playing non-preloaded sound";
-        const sound2 = text.loadAndPlaySound(`${this.baseUrl}/testsound2.wav`,
+        const soundInstance2 = text.loadAndPlaySound(`${this.baseUrl}/testsound2.wav`,
             {},
             {
                 volume: 0.5
             },
             41);
         await delay(3 * 1000);
-
+        soundInstance2.value.stop();
+        await delay(1 * 1000);
         text.text.contents = "resuming";
-        MRESDK.Sound.resume(instanceHandle);
+        soundInstance.value.resume();
         await delay(3 * 1000);
 
         text.text.contents = "Stopping Sound";
-        MRESDK.Sound.stop(instanceHandle);
+        soundInstance.value.stop();
+        this.app.context.assetManager.unloadSoundAsset(soundAssetPromise.value);
+
         await delay(6 * 1000);
 
         return true;
--- a/packages/sdk/src/types/network/payloads/assets.ts
+++ b/packages/sdk/src/types/network/payloads/assets.ts
@@ -10,6 +10,12 @@ import { AssetLike, AssetSource } from '../../runtime/assets';
 export type CreateColliderType = ColliderType | 'none';
 
 /** @hidden */
+export type LoadSound = Payload & {
+    type: 'load-sound';
+    uri: string;
+};
+
+/** @hidden */
 export type LoadAssets = Payload & {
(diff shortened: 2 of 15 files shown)


packages/sdk/src/types/runtime/assets/assetManager.ts
```diff
@@ -94,6 +97,71 @@ export class AssetManager {
...
+    public loadSoundAsset(groupName: string, uri: string, options: LoadSoundOptions): ForwardPromise<string> {
+
+        log.error('app', "STARTING LOADSOUNDASSET");
+        const soundAssetId = UUID();
+        this.enqueueLoadSoundPromise(
+            soundAssetId, {
+                resolve: () => { /* empty */ },
+                reject: () => { /* empty */ },
+            });
+        const promise = this.sendLoadAssetsPayload({
```
eanders-MS: >sendLoadAssetsPayload [](start = 29, length = 21)

This should utilize one of the existing asset messages, `load-assets` or `create-assets`. I don't think you need the `load-sound` message.

Possibly impacting: Looks like the `load-assets` codepath needs some adjustment to return a ForwardPromise<AssetGroup>.

packages/sdk/src/types/network/payloads/payloads.ts
```diff
@@ -358,6 +360,17 @@ export type SyncAnimations = Payload & {
     animationStates: SetAnimationState[];
 };
 
+/**
+ * @hidden
+ * App to engine. Starts playing a sound.
+ */
+export type PlaySound = Payload & {
```
eanders-MS: >PlaySound [](start = 12, length = 9)

A SetSoundState message would be more flexible, allowing you to play/stop, set pitch, etc... all with one message.

packages/sdk/src/types/runtime/assets/assetManager.ts
```diff
@@ -94,6 +97,71 @@ export class AssetManager {
...
+        if (!this.loadSoundPromises) {
+            this.loadSoundPromises = {};
+        }
+        if (!this.loadSoundPromises[soundAssetId]) {
+            this.loadSoundPromises[soundAssetId] = [];
+        }
+        this.loadSoundPromises[soundAssetId].push(promise);
+    }
+
+    public SoundLoaded(soundAssetId: string): Promise<void> {
```
eanders-MS: >SoundLoaded [](start = 11, length = 11)

This method is not static, so it's name should not start with an upper case letter.

eanders-MS: >SoundLoaded [](start = 11, length = 11)

A better API design would be for AssetManager to have a single `loaded` method that supports all asset types. Will require some structural changes to AssetManager itself.

packages/sdk/src/types/runtime/sound.ts
```diff
@@ -0,0 +1,27 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { SetSoundStateOptions } from '../../sound';
+
+export class SoundInstance {
```
eanders-MS: >SoundInstance [](start = 13, length = 13)

I have a hunch SoundInstance will need a reference to either the AssetManager or the Context so that it can send the `set-sound-state` message (my recommended replacement for the 'play-sound' message).

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -437,6 +439,59 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+        return;
+    }
+
+    /**
+     * Loads a sound file and starts playing.
+     * @param uri Path to a sound asset that needs to be loaded and played.
+     * @param options Adjustments to pitch and volume, and other characteristics.
+     * @param startTimeOffset How many seconds to offset into the sound
+     */
+    public loadAndPlaySound(
```
eanders-MS: >loadAndPlaySound [](start = 11, length = 16)

This might be cleaner as a `playSound` overload rather than a separate function.

packages/sdk/src/types/network/payloads/assets.ts
```diff
@@ -9,6 +9,13 @@ import { AssetLike, AssetSource } from '../../runtime/assets';
 
 export type CreateColliderType = ColliderType | 'none';
 
+/** @hidden */
+export type LoadSound = Payload & {
```
eanders-MS: >LoadSound [](start = 12, length = 9)

I don't think this message is necessary. See related comments in assetManager.ts

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -437,6 +439,59 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+     * Starts playing a preloaded sound.
+     * @param soundAssetId Name of sound asset preloaded using AssetManager.
+     * @param options Adjustments to pitch and volume, and other characteristics.
+     * @param startTimeOffset How many seconds to offset into the sound
+     */
+    public playSound(
+        soundAssetId: string, options: SetSoundStateOptions,
+        startTimeOffset?: number): ForwardPromise<SoundInstance> {
+
+        return createForwardPromise(new SoundInstance("SoundAssetHandle"),
```
eanders-MS: >createForwardPromise [](start = 15, length = 20)

All the work this method is doing should be done in the internal Context class. We don't send messages from here, but always send them from methods on internal/context.ts.

packages/sdk/src/types/runtime/assets/assetManager.ts
stevenvergenz: The trouble is that `load-assets` can return zero or more assets. We can't safely pre-generate IDs for the assets. That said, `createMaterial` and `createTexture` *do* return ForwardPromises, so adding a `createSound` would be consistent.

---
In reply to: [257468876](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/204#discussion_r257468876) [](ancestors = 257468876)

packages/functional-tests/src/tests/sound-test.ts
```diff
@@ -0,0 +1,105 @@
...
+        super(app);
+    }
+
+    public async run(): Promise<boolean> {
+        let success = true;
+
+        success = success && await this.runSoundTest();
+
+        return success;
+    }
```
stevenvergenz: I see you using this pattern a lot in functional tests, is this code doing anything at all? Seems like all it does is return the value from the actual test runner.

packages/functional-tests/src/tests/sound-test.ts
```diff
@@ -0,0 +1,105 @@
...
+                text: {
+                    anchor: MRESDK.TextAnchorLocation.MiddleCenter,
+                    color: { r: 30 / 255, g: 206 / 255, b: 213 / 255 },
+                    height: 0.3
+                }
+            }
+        });
+
+        const text = textPromise.value;
+        await textPromise;
```
stevenvergenz: There's no need to save the promise reference. Can replace with `const text = await Actor.CreateEmpty(...)`

packages/sdk/src/types/network/payloads/assets.ts
stevenvergenz: Agreed. Use the CreateAsset payload below.

---
In reply to: [257468904](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/204#discussion_r257468904) [](ancestors = 257468904)

packages/sdk/src/types/runtime/assets/assetManager.ts
stevenvergenz: The `create-asset` payload was designed for this use case.

---
In reply to: [258148811](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/204#discussion_r258148811) [](ancestors = 258148811,257468876)

stevenvergenz: In fact, I think sound should *be* an asset unless there's compelling reason not to. There's a lot of code here that seems redundant with the existing asset pipeline (unload in particular).

---
In reply to: [258156474](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/204#discussion_r258156474) [](ancestors = 258156474,258148811,257468876)

packages/sdk/src/types/runtime/assets/assetManager.ts
```diff
@@ -94,6 +97,71 @@ export class AssetManager {
...
+
+    public enqueueLoadSoundPromise(soundAssetId: string, promise: ExportedPromise): void {
+        if (!this.loadSoundPromises) {
+            this.loadSoundPromises = {};
+        }
+        if (!this.loadSoundPromises[soundAssetId]) {
+            this.loadSoundPromises[soundAssetId] = [];
+        }
+        this.loadSoundPromises[soundAssetId].push(promise);
+    }
```
stevenvergenz: What is the purpose of this API? The load call already returns a promise, do we need alternative mechanisms for waiting for the load? At a minimum, I think these should be private, as ExportedPromise is ironically not a package-exported type.

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: >loadAndPlaySound [](start = 11, length = 16)

Is this the actual implementation? I don't think this does anything.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -437,6 +439,59 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
         return null;
     }
 
+    /**
+     * Starts playing a preloaded sound.
+     * @param soundAssetId Name of sound asset preloaded using AssetManager.
+     * @param options Adjustments to pitch and volume, and other characteristics.
+     * @param startTimeOffset How many seconds to offset into the sound
+     */
+    public playSound(
```
stevenvergenz: >playSound [](start = 11, length = 9)

This API should follow the pattern of enableText, enableRigidBody, etc.: returns void, but sets an actor property for retrieval afterward.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -437,6 +439,59 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+     * @param startTimeOffset How many seconds to offset into the sound
+     */
+    public loadAndPlaySound(
+        uri: string, loadOptions: LoadSoundOptions,
+        playOptions: SetSoundStateOptions,
+        startTimeOffset?: number): ForwardPromise<SoundInstance> {
+        return createForwardPromise(new SoundInstance("SoundAssetHandle"),
+            new Promise<SoundInstance>((resolve, reject) => {
+                resolve();
+            })
```
stevenvergenz: Could be replaced with `Promise.resolve(soundInstance)`

packages/sdk/src/types/runtime/actor.ts
sorenhan: I wanted to clarify that this was a non-instant function, but I guess simply because we just use a URI, it may be is obvious enough for users.

---
In reply to: [257468901](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/204#discussion_r257468901) [](ancestors = 257468901)

packages/functional-tests/src/tests/sound-test.ts
sorenhan: dunno, I copied and pasted. If you change for other tests then I will adapt for this

---
In reply to: [258149899](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/204#discussion_r258149899) [](ancestors = 258149899)

packages/sdk/src/types/runtime/actor.ts
sorenhan: Correct, this does nothing at the moment - implementation coming later

---
In reply to: [258162529](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/204#discussion_r258162529) [](ancestors = 258162529)

packages/sdk/src/types/runtime/actor.ts
sorenhan: this will be completely rewritten, this does nothing

---
In reply to: [258164784](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/204#discussion_r258164784) [](ancestors = 258164784)

packages/sdk/src/types/network/payloads/payloads.ts
sorenhan: Agree. I'll have to add a flag for play/pause separate from the flag for stop, but that's fine.

packages/sdk/src/types/runtime/sound.ts
sorenhan: yep

packages/sdk/src/types/network/payloads/assets.ts
sorenhan: I'll look at this. At the same time I will try to see if we can come up with a better name - I really don't like the "Create" term for loaded assets - that just feels weird.

packages/sdk/src/types/runtime/assets/assetManager.ts
sorenhan: Yeah, a sound should be an asset. My main concern was it seems like unnecessary overhead per sound to have an assetgroup, but maybe that's not a big deal.

packages/functional-tests/src/tests/sound-test.ts
sorenhan: The goal is to remove "await", but there was a bug before where not everything worked unless you awaited. We should go back and remove unnecessary awaits in the samples instead. Making a task

packages/sdk/src/types/runtime/assets/assetManager.ts
sorenhan: Yes, I believe this is needed. I can make it private, though - good call. The purpose is that while we return a promise that the user can wait on, the user is never required to actually await. If they don't await, the node side will internally wait for the sound to be created before it starts to play. This is similar to the animation system, which internally has to wait on animations to be created before they start animating.

It is different from the material/texture code you wrote, which allows for the material id to be created and assigned right away.

packages/sdk/src/types/runtime/actor.ts
sorenhan: No, because it doesn't match a regular component structure.
1. Sounds (and particle effects) auto-expire (if they are non-looping, of course). The expiration is automatically dealt with by clients, and therefore the node-side can just ignore the results. There's not a good use case for listening to when sounds start/stop, so it would just be unnecessary overhead to track this on node side.
2. There can be multiple simultaneously active sound instances on a single actor

sorenhan:
Closing PR as I have received the feedback I need. Implementation work is ongoing

PR #206, opened 2019-02-19
[SDK] Standardize on lower-case, dash-formatted enum values.

eanders-MS:
This is a cleanup pass on our enum values to use a dash-formatted, lower-case format. For the sake of back-compatibility, I won't merge this PR until https://github.com/Microsoft/mixed-reality-extension-unity/pull/33 has been integrated into AltspaceVR and released.

### DIFF
--- a/packages/sdk/src/types/primitiveTypes.ts
+++ b/packages/sdk/src/types/primitiveTypes.ts
@@ -49,14 +49,14 @@ export enum PrimitiveShape {
      * [[PrimitiveDefinition.uSegments]], and vertical segment count [[PrimitiveDefinition.vSegments]], with normals
      * pointed inward, centered at the origin.
      */
-    InnerSphere = 'innersphere'
+    InnerSphere = 'inner-sphere'
 }
 
 /**
  * The size, shape, and description of a primitive.
  */
 export interface PrimitiveDefinition {
-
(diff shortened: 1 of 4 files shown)


eanders-MS:
I left gltf-gen alone. It's kind of its own thing.

tombuMS:
Curious why we are changing these over to dash-formatted for enum values.  The camel case ones should parse naturally in to the camel case versions in unity without any conversion.  Seems weird that we are requiring the use of a converter in C# when this would just happen automatically if they were camel case.

eanders-MS:
Actually, we were already using a default enum converter on the C# side: the StringToEnumConverter. I replaced that with the DashFormattedEnumConverter.

---
In reply to: [466138145](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/206#issuecomment-466138145) [](ancestors = 466138145)

tombuMS:
> Actually, we were already using a default enum converter on the C# side: the StringToEnumConverter. I replaced that with the DashFormattedEnumConverter.
> 
> In reply to: 466138145 [](ancestors = 466138145)

That is the built in one that performs the automatic conversion in newtonsoft that I was talking about originally.  I should have been more specific in the use of a custom converter for enums.

PR #208, opened 2019-02-19
[DO NOT MERGE] Functional test runner refactor

stevenvergenz:
A new test runner setup. Please ignore changes to the individual tests, they're not ported yet.

image

New test paradigm:
* Tests run indefinitely, until stopped by the runner. This means that state-cycling tests should reset at the end of their sequence.
* Tests are expected to not exceed the volume of 2x1x2 meters around the origin.
* Tests should avoid spawning labels to document state changes. Instead, use `app.setOverrideText`.


eanders-MS:
>         const updateLabel = (lightType: string) => {

Light-test updates its message label. Is this something `expectedResultDescription` could support?

---
Refers to: packages/functional-tests/src/tests/light-test.ts:41 in 106c134. [](commit_id = 106c134066caa338e7139ea20b665e95d3d44220, deletion_comment = False)

eanders-MS:
>     private running = true;

Some of the tests I've written have this `running` field that gets checked periodically to see if the test should continue or not. Is there something in the test base class that could be queried instead?

---
Refers to: packages/functional-tests/src/tests/interpolation-test.ts:12 in 106c134. [](commit_id = 106c134066caa338e7139ea20b665e95d3d44220, deletion_comment = False)

packages/functional-tests/tsconfig.json
```diff
@@ -1,8 +1,18 @@
 {
   "extends": "../tsconfig.packages.json",
-  "include": [
-    "./src"
+  "files": [
```
eanders-MS: Why this change? I want to avoid the use of the "files" section. It's one more thing to maintain. Was "include": [ "./src" ] not working for you?

packages/functional-tests/src/tests/failure-test.ts
```diff
@@ -0,0 +1,14 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { Test } from '../test';
+
+export default class FailureTest extends Test {
+    public expectedResultDescription = "Fails";
```
eanders-MS: Name seems overly specific. How about just `statusMessage`?

packages/functional-tests/tsconfig.json
stevenvergenz: This is strictly temporary. I didn't want it building files I hadn't fixed yet.

stevenvergenz:
>         const updateLabel = (lightType: string) => {

Yes, `this.app.setOverrideText()`

---
In reply to: [465689656](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/208#issuecomment-465689656) [](ancestors = 465689656)

---
Refers to: packages/functional-tests/src/tests/light-test.ts:41 in edfeb35. [](commit_id = edfeb359510a839091b65f095487893aed44f3ec, deletion_comment = False)

stevenvergenz:
>     private running = true;

Yes, those can be replaced with `this.stopped` now. I'm still working on updating all the tests.

---
In reply to: [465690010](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/208#issuecomment-465690010) [](ancestors = 465690010)

---
Refers to: packages/functional-tests/src/tests/interpolation-test.ts:12 in edfeb35. [](commit_id = edfeb359510a839091b65f095487893aed44f3ec, deletion_comment = False)

packages/functional-tests/src/app.ts
```diff
@@ -3,203 +3,252 @@
...
-    private activeTests: { [id: string]: Test } = {};
     private _rpc: MRERPC.ContextRPC;
-    private firstUser: MRESDK.User;
-    private _connectedUsers: { [id: string]: MRESDK.User } = {};
+    private firstUser: MRE.User;
+    private _connectedUsers: { [id: string]: MRE.User } = {};
+    private testNames = Object.keys(Factories).sort();
+    private testResults: { [name: string]: boolean } = {};
+
+    private activeTest: Test = null;
```
eanders-MS: >activeTest [](start = 12, length = 10)

As we discussed offline: When the functional-test app is restarted and reloads into the scene, we need to be able to immediately click on the same point in space to relaunch the same test as before the restart. This is important for rapid dev/test iteration when developing a test. Is that still supported?

stevenvergenz: The query argument `test` will skip to that item in the menu, so yes. There are also `autorun` and `nomenu` arguments available.

eanders-MS: That isn't the same. I don't want the test to auto start.

---
In reply to: [258618909](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/208#discussion_r258618909) [](ancestors = 258618909)

eanders-MS: In a single session, we must be able to:
- click on the same spot in space to start the test after restarting the app.
- explicitly start the test (not auto start).
- run the test multiple times.

---
In reply to: [258619934](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/208#discussion_r258619934) [](ancestors = 258619934,258618909)

stevenvergenz: Maybe I need button labels. Set your URI to `ws://localhost:3901/?test=my-test`, and the app will start on your test, ready to go when you click Start (the bottom button).

---
In reply to: [258620721](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/208#discussion_r258620721) [](ancestors = 258620721,258619934,258618909)

stevenvergenz:
Per our face-to-face meeting, I will redesign the menu to allow skipping to an arbitrary test in one or two clicks. The challenge will be keeping it compact. Will experiment.

PR #211, opened 2019-02-25
[SDK] Avatar attachments

eanders-MS:
Adds the `attachment` actor component. Attachment state changes are sent in the actor update patch and applied client side.

#### Not done yet
- Attachments functional test. Will come as a separate PR.


packages/sdk/src/types/runtime/actor.ts
```diff
@@ -83,6 +87,7 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
     private _rigidBody?: RigidBody;
     private _collider?: Collider;
     private _text?: Text;
+    private _attachment: Attachment;
     private _materialId = ZeroGuid;
```
sorenhan: As we talked about, make this nullable/lazily allocated

PR #213, opened 2019-02-25
Enabled the set collider API on an actor

tombuMS:
Updated the API for enabling colliders to use the actor patching mechanism instead of the separate payload for enabling a collider.

This adds two overloaded functions on the actor for setting a box or sphere collider:

`public setCollider(
        colliderType: 'sphere',
        // collisionLayer: CollisionLayer,
        isTrigger: boolean,
        center?: Vector3Like,
        radius?: number): void;`

`public setCollider(
        colliderType: 'box',
        // collisionLayer: CollisionLayer,
        isTrigger: boolean,
        center?: Vector3Like,
        size?: Vector3Like): void;`


packages/sdk/src/types/runtime/actor.ts
```diff
@@ -694,4 +684,50 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+        }
+    }
+
+    private _setCollider(collider: Partial<ColliderLike>) {
+        if (this._collider) {
+            unobserve(this._collider);
+            this._collider = undefined;
+        }
+
+        if (!this._collider) {
```
eanders-MS: >if (!this._collider) { [](start = 8, length = 22)

I think this check is unnecessary. this._collider will always be undefined here.

packages/sdk/src/types/runtime/collider.ts
```diff
@@ -3,54 +3,64 @@
...
-    constructor(private $owner: Actor) { }
-
-    public copy(from: Partial<ColliderLike>): this {
-        if (!from) return this;
-        if (from.enabled !== undefined) this.enabled = from.enabled;
-        if (from.isTrigger !== undefined) this.isTrigger = from.isTrigger;
-        if (from.collisionLayer !== undefined) this.collisionLayer = from.collisionLayer;
-        if (from.colliderParams !== undefined) this._colliderParams = from.colliderParams;
-        return this;
+    constructor(private $owner: Actor, initFrom: Partial<ColliderLike>) {
```
eanders-MS: >constructor [](start = 4, length = 11)

Do you want to mark the constructor @hidden? I think we want it excluded from documentation.

packages/sdk/src/types/runtime/actor.ts
tombuMS: Yeah.  That was a left over check from before changing the enable to set paradigm.  Removing it.

packages/sdk/src/types/runtime/collider.ts
tombuMS: Yep. Added.

PR #222, opened 2019-02-28
Add warning for setting behavior on actor without a collider.

tombuMS:


packages/sdk/src/types/runtime/actor.ts
```diff
@@ -450,6 +450,11 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
      * @param behavior The type of behavior to set. Pass null to clear the behavior.
      */
     public setBehavior<BehaviorT extends Behavior>(behavior: { new(): BehaviorT }): BehaviorT {
+        if (!Collider) {
```
eanders-MS: Does this compile? I think it should be lower case.

eanders-MS: this.collider

tombuMS: It does compile as it is an import symbol, but it is the incorrect one.  Fixed!

PR #225, opened 2019-03-01
[SDK] LookAt, take 3

### LINKED ISSUE #140: LookAt isn't synchronized
Look-at functional tests (and functional tests that align the label towards user) only look proper on the first user's screen. Expected behavior: on all machines it will align towards the same user (i.e. it is not supposed to align towards user A on machine A and user B on machine B)

eanders-MS:
This implements LookAt as an actor component. The component has three fields:
- actorId: The actor to face toward.
- mode: The LookAtMode (unchanged from before).
- backward: Look away from the target rather than toward.

Actors can face other actors, but can no longer face users. At least not directly. Facing a user is achieved by attaching an empty object to the user and looking at that instead.

I chose to not support offset rotation. Offset rotation can be achieved by adding an empty object at an offset from the actor you're appearing to face. Bonus: That empty object can be animated for an interesting effect.

Fixes #140


davidjcalabrese:
What is the use case for LookAt?

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -384,6 +389,44 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+        let actorId = ZeroGuid;
+        if (typeof (actorOrActorId) === 'object' && actorOrActorId.id !== undefined) {
+            actorId = actorOrActorId.id;
+        } else if (typeof (actorOrActorId) === 'string') {
+            actorId = actorOrActorId;
+        }
+        // Allocate component if necessary.
+        if (!this._lookAt) {
+            this._lookAt = new LookAt();
+            // Set up observed properties in the next tick, giving us a chance to initialize the values first (below).
```
stevenvergenz: >// Set up observed properties in the next tick, giving us a chance to initialize the values first (below). [](start = 12, length = 106)

I think duplicating the copy call for the initialization wouldn't be as bad as delaying the observation. That delay seems really strange.

packages/sdk/src/types/runtime/lookAt.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { LookAtMode, Vector3, Vector3Like } from "../..";
+import { ZeroGuid } from "../../constants";
+
+export interface LookAtLike {
+    actorId: string;
+    mode: LookAtMode;
+    backward: boolean;
```
stevenvergenz: >backward: boolean; [](start = 4, length = 18)

Interesting, a backward field? Not using the transform rotation as an offset?

packages/sdk/src/types/runtime/actor.ts
eanders-MS: Agreed. I tried this a few different ways. Let me update to a better one. Once we're settled on a good pattern, I'll update the other enable* calls to match.

stevenvergenz:
Oftentimes we want text labels to face the user. This is much cheaper than setting the rotation every frame from Node.

---
In reply to: [468493884](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/225#issuecomment-468493884) [](ancestors = 468493884)

eanders-MS:
@davidjcalabrese It's for when you want stuff to look at other stuff :). A concrete example: You have a pet that communicates with body language that it is hungry by looking at your avatar's hand, then face, then hand. LookAt is one of a set of features that aim to mitigate latency issues stemming from the fact that the MRE app is running remotely. InterpolateTo is another example. These behaviors could be implemented server side, but it would be working on out of date information, and updates would have a noticeable delay. Implementing it in-engine provides a superior experience..

packages/sdk/src/types/runtime/lookAt.ts
eanders-MS: Not sure using the transform rotation as an offset is a good idea, since any original rotation would cause LookAt to not look at, and that could feel buggy. After rotating an object, you would have to know to zero out the rotation to make LookAt work again. We could send an offset rotation parameter instead. Backward is really just a 180 rotation offset.

eanders-MS: It would be reasonable to support 90 offset, etc. I think an offset rotation parameter will be better.

eanders-MS: Going to stick with `backward`, for simplicity. Other rotations can be achieved using an empty object.

---
In reply to: [261451716](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/225#discussion_r261451716) [](ancestors = 261451716)

packages/sdk/src/types/runtime/actor.ts
eanders-MS: Removed call to nextTick.

---
In reply to: [261443428](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/225#discussion_r261443428) [](ancestors = 261443428)

PR #231, opened 2019-03-02
Multiple fixes for issues uncovered while running functional tests

eanders-MS:
**Fixes**
- AnimateTo should not have a reply message. This was causing the app to hang when one user left.
- readPath (in patching system) should not die on undefined values. Now it just ignores them.
- During client sync, we don't have a reliable way to know whether to send reply messages back to the app, so now we just send them all back and let the app's protocol layer sort it out.

**Unresolved issues**
- When a client disconnects, we don't clean up outstanding promises for messages awaiting replies. This will cause a timeout and kill the session.
- When the session is killed this way, it doesn't clean up properly.
- I will follow up on these Monday.

### DIFF
--- a/packages/sdk/src/protocols/protocol.ts
+++ b/packages/sdk/src/protocols/protocol.ts
@@ -105,7 +105,7 @@ export class Protocol extends EventEmitter {
             if (this.timeoutSeconds > 0) {
                 return setTimeout(() => {
                     // tslint:disable-next-line:max-line-length
-                    const reason = `Timed out awaiting response for ${message.payload.type}, id:${message.id}.`;
+                    const reason = `${this.name} timed out awaiting response for ${message.payload.type}, id:${message.id}.`;
                     log.error('network', reason);
                     this.rejectPromiseForMessage(message.id, reason);
                     this.conn.close();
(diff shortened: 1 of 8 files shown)


packages/sdk/src/protocols/protocol.ts
```diff
@@ -198,9 +198,14 @@ export class Protocol extends EventEmitter {
...
         const queuedPromise = this.promises[message.replyToId];
+        // The way join-synchronization works, the app may recieve the occasional unexpected reply message.
+        // It is harmless; something we can live with for now. We do want to know if we reeive an unexpected reply
+        // elsewhere, as that could be a sign of a protocol design issue.
         if (!queuedPromise) {
-            // tslint:disable-next-line:no-console max-line-length
-            console.error(`[ERROR] ${this.name} received unexpected reply message! payload: ${message.payload.type}, replyToId: ${message.replyToId}`);
+            if (this.name !== "Execution") {
+                // tslint:disable-next-line:no-console max-line-length
+                console.error(`[ERROR] ${this.name} received unexpected reply message! payload: ${message.payload.type}, replyToId: ${message.replyToId}`);
```
tombuMS: Is there a reason that this is console.error and not through our normal logging mechanism?

eanders-MS: Until very recently we didn't have app logging enabled by default on our samples and functional tests, and I really want to know if this condition was ever hit, so I had it go through the console. It could be switched over to app logging now. There are a few other places in the code where we're logging to console. I'll clean them up in a separate PR.

PR #233, opened 2019-03-02
[NOT FOR MERGE] Grab Behavior API v1

tombuMS:
This is the API design proposal for the v1 of grab behavior associated with actors for the TS side of the SDK.

### DIFF
--- a/packages/sdk/src/types/internal/context.ts
+++ b/packages/sdk/src/types/internal/context.ts
@@ -48,7 +48,6 @@ import {
     SetBehavior,
     SetSoundState,
     UpdateCollisionEventSubscriptions,
-    UpdateGrabbable,
     UpdateSubscriptions,
 } from '../network/payloads';
 
@@ -600,21 +599,6 @@ export class InternalContext {
                 actorId,
                 behaviorType: newBehaviorType || 'none'
             } as SetBehavior);
-        }
-    }
-
-    public setGrabbability(actorId: string, grabbable: boolean) {
-        const actor = this.actorSet[actorId];
-        if (actor) {
-            actor.created().then(() => {
-                this.protocol.sendPayload({
-                    type: 'update-grabbable',
-                    actorId,
-                    grabbable
-                } as UpdateGrabbable);
-            }).catch((reason: any) => {
-                log.error('app', `Failed to update grabbability on actor ${actor.id}.`, reason);
-            });
         }
     }
 }
--- a/packages/sdk/src/types/internal/context.ts
+++ b/packages/sdk/src/types/internal/context.ts
@@ -48,6 +48,7 @@ import {
     SetBehavior,
     SetSoundState,
     UpdateCollisionEventSubscriptions,
+    UpdateGrabbable,
     UpdateSubscriptions,
 } from '../network/payloads';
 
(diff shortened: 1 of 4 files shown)


packages/sdk/src/types/network/payloads/payloads.ts
```diff
@@ -317,6 +317,17 @@ export type SetBehavior = Payload & {
...
+ * @hidden
+ * App to engine.  Update whether the actor is grabbable
+ * or not.
+ */
+export type UpdateGrabbable = Payload & {
+    type: 'update-grabbable';
+    actorId: string;
+    grabbable: boolean;
+};
+
```
tombuMS: In thinking about this some more, I think that this will not be needed, as the grabbable state should be able to be updated through patching alone, but will investigate this further in the Unity runtime to be sure. Will likely be removed.

packages/sdk/src/types/internal/context.ts
```diff
@@ -738,4 +659,19 @@ export class InternalContext {
...
+    public setGrabbability(actorId: string, grabbable: boolean) {
+        const actor = this.actorSet[actorId];
+        if (actor) {
+            actor.created().then(() => {
+                this.protocol.sendPayload({
+                    type: 'update-grabbable',
+                    actorId,
+                    grabbable
+                } as UpdateGrabbable);
+            }).catch((reason: any) => {
```
tombuMS: This implementation may change away from a second payload.  See below comment about investigating that within the unity runtime, as behaviors are kind of a special case.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -97,6 +97,8 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
     private _attachment: Attachment;
     private _lookAt: LookAt;
     private _materialId = ZeroGuid;
+    private _grabbable = false;
+    private _grab = new DiscreteAction();
```
tombuMS: Will be moving the _grab out of this area and prefixing with $ as it should not be observed.

packages/sdk/src/types/internal/context.ts
eanders-MS: This should utilize the actor patch. No need for a new payload type.

---
In reply to: [261903047](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/233#discussion_r261903047) [](ancestors = 261903047)

packages/sdk/src/types/network/payloads/payloads.ts
eanders-MS: Yep I think this should be done via actor patch.

---
In reply to: [261902987](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/233#discussion_r261902987) [](ancestors = 261902987)

packages/sdk/src/types/runtime/actor.ts
eanders-MS: No need for that since we don't observe the entire actor, just parts of it.

---
In reply to: [261903170](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/233#discussion_r261903170) [](ancestors = 261903170)

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -148,6 +151,14 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
         this.actorChanged('materialId');
     }
 
+    public get grabbable() { return this._grabbable; }
+    public set grabbable(value) {
+        if (value !== this._grabbable) {
+            this._grabbable = value;
+            this.actorChanged('grabbable');
+        }
```
eanders-MS: Looks good.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -63,6 +63,7 @@ export interface ActorLike {
     attachment: Partial<AttachmentLike>;
     lookAt: Partial<LookAtLike>;
     materialId: string;
+    grabbable: boolean;
```
stevenvergenz: >grabbable: boolean; [](start = 4, length = 19)

Why is grabbable part of an actor patch, but not buttonlike?

stevenvergenz:
I'm confused. Is grabbable not using the Behavior system? It seems like everything is right on the actor in this PR.

tombuMS:
Closing and putting out the PR for merging.

PR #243, opened 2019-03-06
functional test polish

sorenhan:
changed FT orientation to be visible without rotating space component
Made all items above ground (y>0) and in front of wall (z<0).
added ground and wall geometry
moved everything to be within (or nearly within) the floor area.
added lights to some scenes that were hard to see
moved menu to the wall, and stop/return buttons to the front

### DIFF
--- a/packages/functional-tests/src/tests/input-test.ts
+++ b/packages/functional-tests/src/tests/input-test.ts
@@ -16,6 +16,21 @@ export default class InputTest extends Test {
     private model: MRE.Actor;
 
     public async run(): Promise<boolean> {
+        MRE.Actor.CreateEmpty(this.app.context, {
+            actor: {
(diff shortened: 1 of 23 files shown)


packages/functional-tests/src/tests/input-test.ts
```diff
@@ -26,10 +41,12 @@ export default class InputTest extends Test {
             actor: {
                 name: 'clickable',
                 transform: {
-                    scale: { x: 0.4, y: 0.4, z: 0.4 }
+                    scale: { x: 0.4, y: 0.4, z: 0.4 },
+                    position: { x: 0, y: 1, z: -1 }
                 }
             }
         }).value;
+        await this.model;
```
eanders-MS: Incorrect to wait on this.

PR #260, opened 2019-03-14
[SDK]: Grab functionality added to actor

tombuMS:
Grab is being added to the actors within the sdk as a property on the actor called `grabbable`.  By setting this field, a developer can enable/disable grab on an actor.  Additionally there is now a method on actor that allows a developer to set a handler to be called when a grab begins and/or ends.  This is the new method is `onGrab(grabState: 'begin' | 'end', handler: ActionHandler)`.

Unity runtime reference: [PR#58](https://github.com/Microsoft/mixed-reality-extension-unity/pull/58)


packages/sdk/src/types/runtime/actor.ts
```diff
@@ -98,8 +99,11 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
     private _text: Text;
     private _attachment: Attachment;
     private _lookAt: LookAt;
+    private _grabbable = false;
     // tslint:enable:variable-name
 
+    private grab = new DiscreteAction();
```
eanders-MS: >grab [](start = 12, length = 4)

Most objects won't be grabbable. Can `grab` be lazily allocated?

tombuMS: I don't see why not, but we don't actually do that anywhere else in the behavior system at the moment.

tombuMS: I don't see why not, but we don't actually do that anywhere else in the behavior system at the moment.

eanders-MS: Curious: Would `grab` become a component once we have that system in place?

---
In reply to: [266113938](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/260#discussion_r266113938) [](ancestors = 266113938)

tombuMS: I don't know.  My initial gut reaction would be possibly, but it is more like a property of the actor.  Not sure at this moment.  I can see arguments for either way.

packages/functional-tests/src/tests/grab-test.ts
```diff
@@ -0,0 +1,143 @@
...
+            this.state = 1;
+            this.cycleState();
+        });
+        this.model.onGrab('end', _ => {
+            this.state = 2;
+            this.cycleState();
+        });
+
+        // Create two grabbable cubes that can be played with at will.  Position left
+        // anr right of the monkey.
```
eanders-MS: >anr [](start = 11, length = 3)

anr -> and

PR #269, opened 2019-03-18
[DO NOT MERGE] Player masking API

stevenvergenz:


packages/sdk/src/types/runtime/appearance.ts
```diff
@@ -19,13 +20,22 @@ export class Appearance implements AppearanceLike {
     private _materialId = ZeroGuid;
     // tslint:enable:variable-name
 
-    /** Whether the actor with this appearance should be visible */
+    /**
+     * This actor's visibility preference, independent of its parent. See
+     * [[activeAndEnabled]] for the actual visibility state.
+     */
     public enabled = true;
```
eanders-MS: Seems like `enabled` should be a getter now, so it can return enabledFor's packed value, if it is set.

packages/sdk/src/types/runtime/appearance.ts
```diff
@@ -4,10 +4,11 @@
  */
 
 import { Actor, Material } from '.';
+import UserGroup from '../userGroup';
 import { ZeroGuid } from '../../constants';
 
 export interface AppearanceLike {
-    enabled: boolean;
+    enabled: boolean | number;
```
eanders-MS: Will you be adding this to the behavior system too? Would be good to include that in this PR. Let's get clarity on how UserGroups apply to triggers and user input.

packages/sdk/src/types/userGroup.ts
```diff
@@ -0,0 +1,48 @@
...
+ * [[Appearance.enabled]], and [[Collider.enabled]].
+ */
+export default class UserGroup extends Set<string> {
+    constructor(initialContents: string[] = []) {
+        super(initialContents);
+    }
+
+    public add(item: string): this {
+        if (!groupIndex[item]) {
+            // according to MDN, 2^52 is the largest pow2 that can be stored in a number primitive
```
eanders-MS: I suppose you'll be using a long on the C# side...

packages/sdk/src/types/runtime/user.ts
```diff
@@ -30,11 +36,18 @@ export class User implements UserLike {
     private _properties: { [name: string]: string };
     // tslint:enable:variable-name
 
+    /**
+     * This user's group memberships.
+     */
+    public readonly groups = new UserGroup();
```
eanders-MS: Most users won't have groups, so this can be lazily allocated.

packages/sdk/src/types/runtime/appearance.ts
```diff
@@ -19,13 +20,22 @@ export class Appearance implements AppearanceLike {
...
 
+    /** Whether this actor is visible */
     public get activeAndEnabled() {
         return (!this.actor.parent || this.actor.parent.appearance.enabled) && this.enabled;
     }
 
+    /**
+     * Display this actor only for members of the contained groups.
+     */
+    public readonly enabledFor = new UserGroup();
```
eanders-MS: Most appearance objects won't have group affinities, so this can be lazily allocated.

packages/sdk/src/types/userGroup.ts
```diff
@@ -0,0 +1,48 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+let lastIndex = 0;
```
eanders-MS: We need to avoid global state in the SDK.

packages/sdk/src/types/userGroup.ts
stevenvergenz: `ulong` most likely, yeah. That's the 64-bit one, right?

---
In reply to: [266634369](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/269#discussion_r266634369) [](ancestors = 266634369)

packages/sdk/src/types/runtime/appearance.ts
stevenvergenz: I was not planning on adding this to behaviors, no. Instead I would add it to the colliders.

---
In reply to: [266632591](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/269#discussion_r266632591) [](ancestors = 266632591)

packages/sdk/src/types/runtime/userGroup.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private getOrAddMapping(name: string): number {
+        if (!this.context) {
+            throw new Error("Cannot manipulate a user group without associating it with a context first");
+        }
+
+        const mapping = this.context.internal.userGroupMapping;
+        if (!mapping[name]) {
+            const lastIndex = Object.keys(mapping).length;
```
eanders-MS: >lastIndex [](start = 18, length = 9)

Property order is preserved since ES6, so I think this is safe.

packages/sdk/src/types/runtime/userGroup.ts
```diff
@@ -0,0 +1,73 @@
...
+            // according to MDN, 2^52 is the largest pow2 that can be stored in a number primitive
```
eanders-MS: >according to MDN [](start = 15, length = 16)

Can you link to the docs?

packages/sdk/src/types/runtime/userGroup.ts
```diff
@@ -0,0 +1,73 @@
...
+            if (lastIndex > 52) {
```
eanders-MS: >lastIndex > 52 [](start = 16, length = 14)

If 2^52 is the max bitwise bit, then 1<<51 gets you there, so I think this should be `lastIndex >= 52`.

packages/sdk/src/types/runtime/userGroup.ts
```diff
@@ -0,0 +1,73 @@
...
+    private context: Context;
+
+    constructor(initialContents: Iterable<string> = null, context: Context = null) {
+        super(initialContents);
+        if (context) {
+            this.setContext(context);
+        }
+    }
+
+    public setContext(c: Context) {
```
eanders-MS: >setContext [](start = 11, length = 10)

This needs to be hidden. But, I'm not sure it's even needed. Can we just always pass context to the constructor? When is it desired to not have a context reference here?

packages/sdk/src/types/runtime/userGroup.ts
```diff
@@ -0,0 +1,73 @@
...
+ * A set of user group IDs. User groups are used to selectively enable several different
+ * properties of actors based on the memberships of the viewing user. See [[User.groups]],
+ * [[Appearance.enabled]].
+ */
+export default class UserGroup extends Set<string> {
+    // tslint:disable:no-bitwise
+
+    private context: Context;
+
+    constructor(initialContents: Iterable<string> = null, context: Context = null) {
```
eanders-MS: >context: Context = null [](start = 58, length = 23)

When would you not want to pass a context to the constructor?

packages/sdk/src/types/runtime/user.ts
```diff
@@ -50,6 +68,13 @@ export class User implements UserLike {
         if (from.id !== undefined) this._id = from.id;
         if (from.name !== undefined) this._name = from.name;
         if (from.properties !== undefined) this._properties = from.properties;
+        if (from.groups !== undefined) {
+            if (typeof from.groups === 'number') {
+                this._groups = UserGroup.FromPacked(this._context, from.groups);
+            } else {
+                this._groups = from.groups;
```
sorenhan: I don't understand this bit - is this because we allow to store this both as a class and a number here?

sorenhan:
I like it. Would love to see a functional test - maybe 2 buttons (toggle red group, toggle blue group), and a red and a blue cube, so you can see if single and multiple groups work). I know you're a fan of creating the test first, so I hope that's the next step.

Issue #30, opened 2018-12-06
Feature Request: Player Masking


sorenhan:
Need the ability to enable behaviors for one or all users, or possibly user group/teams.
Need the ability to enable actors for one user, all users, all users except one, or possibly user group/teams.

must be dynamic (as people join/leave teams)
Must support recording+Playback.

some use cases: 
first person shooter - player sees their own first person rig vs everyone else sees a real 3d model, 
when players die, they can see everyone? 
card games - hide cards from anyone but player - but the moment player plays a card, it becomes visible to everyone.

sorenhan:
#137 (single-user actors) would cover the "enable actors for one user" subset of the functionality requested here - especially the " hide cards from anyone but player" scenario

sorenhan:
This should not change how we synchronize actors - these actors are synchronized across all machines still, it's only about disabling rendering and behavior collision for specific users (physics simulation must still be unaffected, and simulation must still happen).

I see 2 separate mechanisms for this: 
1. UserGroup flags - I imagine this as simple bit mask, set for each actor, and for each user - there are 32 bits, representing 32 groups. each actor has an integer for which user groups should include this actor, and it's and'ed with the user's usergroup flag. If there is no overlap between actor's groups and user's groups, then that actor, and all children recursively are invisible and non-collidable for behaviors. 
We would probably have a list of default masks (everyone, spectator, participant, team A, team B, team C), Default values for actors: "all" flag, for users: "all|participant|TeamA" flag.
1b. This may be a slight overload, but it could easily be used for low/medium/high quality settings (auto filled in from altspaceVR settings), if AltspaceVR just auto assigns the user a quality setting (go/gear/Daydream=low, pc discrete=high, pc integrated/Quest/2d mobile=medium)

2. User exclude flag per actor - The actor should also have a "currently excluded" user, which if matching the user, then that actor, and all children recursively are invisible and non-collidable for behaviors. (i.e. follows same rules as when usergroup disables an entity)

sorenhan:
We would *probably* want to have 2 separate integers - one for visibility, and one for behaviors. I can see leaving something visible for everyone but not interactable for everyone.

sorenhan:
There is some overlap in functionality with #30 (as that's also about enabling/disabling different actor components), so plan these together

eanders-MS:
[...]

PR #270, opened 2019-03-20
[SDK] User Masking

stevenvergenz:
## API Changes
* Add a `GroupMask` class that encapsulates group add/remove logic and conversion to/from a bitfield
* Change `Appearance.enabled` from `boolean` to `boolean | GroupMask`
* Add `User.groups: GroupMask` field

## Infrastructure Changes
* Call `toJSON` on objects resolved from `readPath` if they have such a method
* Make `User` patchable, and send corresponding `user-update` messages only to the specified user

## Misc
* Added a functional test demoing user masking


packages/sdk/src/types/runtime/userGroupCollection.ts
```diff
@@ -0,0 +1,139 @@
...
+import { Context } from '.';
+
+// tslint:disable:no-bitwise
+
+/**
+ * A set of user group IDs. User groups are used to selectively enable several different
+ * properties of actors based on the memberships of the viewing user. All users not assigned
+ * a group are in the `default` group. See [[User.groups]], [[Appearance.enabled]].
+ */
+export default class UserGroupCollection extends Set<string> {
```
eanders-MS: Don't export as a default, for consistency with peers at this scope.

packages/sdk/src/types/runtime/user.ts
```diff
@@ -5,11 +5,15 @@
 
 import { Context } from '../..';
 import { InternalUser } from '../internal/user';
+import UserGroupCollection from './userGroupCollection';
```
eanders-MS: This import is unnecessary. Add UserGroupCollection to the first import above.

packages/sdk/src/utils/readPath.ts
```diff
@@ -25,5 +25,5 @@ export default function readPath(src: any, dst: any, ...path: string[]) {
         }
         src = src[field];
     }
-    dst[field] = src;
+    dst[field] = src.toJSON ? src.toJSON() : src;
```
eanders-MS: Nice, but need to add null protection. `src` can be null/undefined/etc.

packages/sdk/src/types/runtime/index.ts
```diff
@@ -18,3 +18,4 @@ export * from './assets';
 export * from './physics';
 export * from './lookAt';
 export * from './appearance';
+export { default as UserGroupCollection } from './userGroupCollection';
```
eanders-MS: Is it necessary to do this differently than all the other exports?

packages/sdk/src/types/runtime/userGroupCollection.ts
```diff
@@ -0,0 +1,139 @@
...
+    public static readonly ALL_PACKED = ~0;
+    public static readonly NONE_PACKED = 0;
+
+    private context: Context;
+
+    constructor(initialContents: Iterable<string> = null, context: Context = null) {
```
eanders-MS: It would be simpler to just always require a context passed to the constructor. Then you could delete setContext entirely.

packages/sdk/src/types/runtime/userGroupCollection.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    private getOrAddMapping(name: string): number {
+        if (!this.context) {
+            throw new Error("Cannot serialize a user group without associating it with a context first");
+        }
+
+        const mapping = this.context.internal.userGroupMapping;
+        if (!mapping[name]) {
+            const lastIndex = Object.keys(mapping).length;
+            // all bitwise inputs are coerced to 32-bit signed ints
```
eanders-MS: If someone runs out of groups we can switch to a solution that uses a growable UInt32Array.

packages/sdk/src/types/internal/context.ts
```diff
@@ -190,6 +193,12 @@ export class InternalContext {
     }
 
     private createActorFromPayload(payload: CreateActorCommon): ForwardPromise<Actor> {
+        // bind all user group mappings before serialization
+        const ug = safeGet(payload, 'actor', 'appearance', 'enabled');
```
eanders-MS: This is brittle. It will be cleaner and more robust to pass context in the constructor.

packages/sdk/src/types/runtime/userGroupCollection.ts
stevenvergenz: My thoughts exactly. Or a byte array stored as a string or something. Lots of scaling options, but my intuition says that 31 groups should be plenty.

---
In reply to: [267573653](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/270#discussion_r267573653) [](ancestors = 267573653)

packages/sdk/src/adapters/multipeer/rules.ts
```diff
@@ -1047,6 +1047,15 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
...
     'user-update': {
-        ...ClientOnlyRule
+        ...DefaultRule,
+        client: {
+            beforeQueueMessageForClient: (
+                session: Session,
+                client: Client,
+                message: Message<Payloads.UserUpdate>
+            ) => {
+                return client.userId === message.payload.user.id ? message : null;
```
eanders-MS: >return client.userId === message.payload.user.id ? message : null; [](start = 16, length = 66)

What are you trying to accomplish here? In this method I think you will need to do something similar to what 'actor-update' is doing above, and this kind of filtering will be done elsewhere.

stevenvergenz: I want this message to only be sent to the client that is hosting the given user. The other clients don't know about other users AFAICT, so wouldn't be able to do anything with this info.

---
In reply to: [267581467](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/270#discussion_r267581467) [](ancestors = 267581467)

stevenvergenz:
Fixes #30

PR #274, opened 2019-03-25
[SDK] User Masking take 2

stevenvergenz:
## API Changes
* Add a `GroupMask` class that encapsulates group add/remove logic and conversion to/from a bitfield
* Change `Appearance.enabled` from `boolean` to `boolean | GroupMask`
* Add `User.groups: GroupMask` field

## Infrastructure Changes
* Call `toJSON` on objects resolved from `readPath` if they have such a method
* Make `User` patchable, and send corresponding `user-update` messages only to the specified user
* Provide a hook to rewrite all actor patch objects leaving the app. `Actor.sanitize` is called by both the create and update flows.

## Misc
* Added a functional test demoing user masking


packages/sdk/src/types/internal/actor.ts
```diff
@@ -64,8 +65,8 @@ export class InternalActor implements InternalPatchable<ActorLike> {
         if (patch) {
             patch.id = this.actor.id;
             delete this.patch;
+            return Actor.sanitize(patch);
```
eanders-MS: >return Actor.sanitize(patch); [](start = 12, length = 29)

This isn't the right place to resolve json values. We can't have unresolved objects in the patch between frames or they might get updated before serialization.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -718,6 +719,20 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+     */
+    public static sanitize(msg: ActorLike): ActorLike;
+    public static sanitize(msg: Partial<ActorLike>): Partial<ActorLike>;
+    public static sanitize(msg: ActorLike | Partial<ActorLike>): ActorLike | Partial<ActorLike> {
+        msg = resolveJsonValues(msg);
+        if (msg.appearance) {
+            msg.appearance = Appearance.sanitize(msg.appearance);
+        }
+        return msg;
+    }
```
eanders-MS: Actor is an odd place for this to live. Can it just be a utility function?

eanders-MS: I guess it's fine here.

---
In reply to: [268831340](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/274#discussion_r268831340) [](ancestors = 268831340)

packages/sdk/src/types/runtime/user.ts
```diff
@@ -3,37 +3,61 @@
...
+    public get groups() {
+        if (!this._groups) {
+            this._groups = new GroupMask(this._context);
+            this._groups.allowDefault = false;
+            this._groups.onChanged(() => this.userChanged('groups'));
+        }
+        return this._groups;
+    }
+    public set groups(val) {
+        this._groups = val;
```
eanders-MS: It seems that multiple actors and users could share one GroupMask instance, but there is only one callback handler on GroupMask. If I created a GroupMask for "red team" and assigned it to multiple users and actors, only one would be notified of group changes.

You could have a private EventEmitter and let it handle all the listeners. You'd need a way to unregister a handler, but EventEmitter can do that too.

packages/sdk/src/types/runtime/groupMask.ts
```diff
@@ -0,0 +1,147 @@
...
+            for (const group of initialContents) {
+                if (this.allowDefault || group !== 'default') {
+                    this.getOrAddMapping(group);
+                    super.add(group);
+                }
+            }
+        }
+    }
+
+    public static All(context: Context) {
```
eanders-MS: Please document all public elements, or mark them @hidden.

packages/sdk/src/types/runtime/appearance.ts
```diff
@@ -3,34 +3,99 @@
...
+    public get enabled(): boolean | GroupMask {
+        if (this.enabledPacked === GroupMask.ALL_PACKED) {
+            return true;
+        } else if (this.enabledPacked === GroupMask.NONE_PACKED) {
+            return false;
+        } else {
+            return this.enabledFor;
+        }
+    }
+    public set enabled(value: boolean | GroupMask) {
```
eanders-MS: I might be mistaken, but I think getters and setters each need their own doc headers. Grouping them won't make the documentation work for both.

packages/sdk/src/types/runtime/user.ts
eanders-MS: Alternately, don't expose GroupMask to users. It can be an internal thing.

stevenvergenz: I think I'd rather enforce single usage of GroupMask objects. Should be pretty easy to clone the mask in the property setters.

---
In reply to: [268855539](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/274#discussion_r268855539) [](ancestors = 268855539)

stevenvergenz: I don't think that's viable. The whole point of this class is to provide a user-friendly frontend for the packed bit masks. How else would they provide group membership?

---
In reply to: [268868693](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/274#discussion_r268868693) [](ancestors = 268868693)

packages/sdk/src/types/runtime/appearance.ts
stevenvergenz: The header is for the property, and they're the same property. I've checked it, this works.

---
In reply to: [268862361](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/274#discussion_r268862361) [](ancestors = 268862361)

packages/sdk/src/types/runtime/groupMask.ts
stevenvergenz: Okay. Should I also document the super-class methods? I don't know how typedoc handles those.

---
In reply to: [268861446](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/274#discussion_r268861446) [](ancestors = 268861446)

packages/sdk/src/types/internal/actor.ts
stevenvergenz: This is really the only place to do it. We can't resolve every time the patch updates, it would produce a ton of garbage objects that get immediately discarded. Is it really so bad if they change? Which objects end up in the patch that don't trigger updates on changed anyway?

---
In reply to: [268830814](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/274#discussion_r268830814) [](ancestors = 268830814)

PR #275, opened 2019-03-25
[NOT FOR MERGE]: Collider events design

tombuMS:
See COLLIDER_EVENTS.md


COLLIDER_EVENTS.md
```diff
@@ -0,0 +1,66 @@
...
+The collision event is raised when two actors that both have colliders on them participate in a collision event, one for when the collision enters and one for when the collision exits..  This collision event will contains data that is passed to the event handlers attached for the collision event and state.
+
+## Collision Data
+
+Collision data is specific to only collision events and contains information on the other actor, as well as physics related information about the collision.
+
+```ts
+export interface ContactPoint {
+    normal: Vector3;
+    point: Vector3;
```
eanders-MS: point and normal will be in application coordinate space, correct?

COLLIDER_EVENTS.md
```diff
@@ -0,0 +1,66 @@
...
+    separation: number;
+}
+
+export interface CollisionData {
+    actorHit?: Actor;
+    contacts: ContactPoint[];
+    impulse: Vector3;
```
eanders-MS: impulse and relativeVelocity also expressed in app coord space.

packages/sdk/src/types/runtime/collider.ts
```diff
@@ -24,8 +30,11 @@ export class Collider implements ColliderLike {
     // public collisionLayer = CollisionLayer.Object;
 
     // Readonly params that are not patchable or observable.
-    // tslint:disable-next-line:variable-name
+    // tslint:disable:variable-name
     private _colliderGeometry: Readonly<ColliderGeometry>;
+    private _collisionHandlers: CollisionHandlers = {};
```
eanders-MS: You can do all this with a single EventEmitter. I think an EventEmitter is more appropriate here, since you can have multiple listeners registered for an event.

packages/sdk/src/types/runtime/collider.ts
```diff
@@ -55,6 +64,58 @@ export class Collider implements ColliderLike {
...
     }
 
+    /**
+     * Add a collision event handler for the given collision event state.
+     * @param state The state of the collision event.
+     * @param handler The handler to call when a collision event with the matching
+     * collision event state is received.
+     */
+    public onCollision(state: CollisionEventState, handler: CollisionHandler) {
+        if (!!this._collisionHandlers[state]) {
```
eanders-MS: I think this logic is inverted.

packages/sdk/src/types/runtime/collider.ts
```diff
@@ -55,6 +64,58 @@ export class Collider implements ColliderLike {
...
+    }
+
+    /**
+     * Add a trigger event handler for the given collision event state.
+     * @param state The state of the trigger event.
+     * @param handler The handler to call when a trigger event with the matching
+     * collision event state is received.
+     */
+    public onTrigger(state: CollisionEventState, handler: CollisionHandler) {
+        if (!!this._triggerHandlers[state]) {
```
eanders-MS: logic is inverted

COLLIDER_EVENTS.md
```diff
@@ -0,0 +1,66 @@
...
+# Collision Event
+
+The collision event is raised when two actors that both have colliders on them participate in a collision event, one for when the collision enters and one for when the collision exits..  This collision event will contains data that is passed to the event handlers attached for the collision event and state.
+
+## Collision Data
+
+Collision data is specific to only collision events and contains information on the other actor, as well as physics related information about the collision.
+
+```ts
+export interface ContactPoint {
```
eanders-MS: And, maybe later we can add the app the actor came from :)

COLLIDER_EVENTS.md
tombuMS: Yes.  They would be converted to app coordinates.

COLLIDER_EVENTS.md
tombuMS: Yep

COLLIDER_EVENTS.md
tombuMS: I am intentionally filtering out actors that are not in the same app for the first pass.  We need an actual design on the inter-app communication.  😃

COLLIDER_EVENTS.md
```diff
@@ -0,0 +1,66 @@
...
+    actorHit?: Actor;
+    contacts: ContactPoint[];
+    impulse: Vector3;
+    relativeVelocity: Vector3;
+}
+```
+
+# Trigger Volume Events
+
+A trigger volume event is raised when an actor that has a collider on it enters or exits the bounds of another actor's collider that is configured to be a trigger. This trigger event will contain data that is passed to the event containining the actor id of the other actor.
```
stevenvergenz: > enters or exits

So triggers are not fired continuously while within the bounds, correct?

tombuMS: Right.  Just enter and exit.

PR #287, opened 2019-04-02
Behavior callbacks pass User objects, not IDs

stevenvergenz:
This PR changes all interfaces into the Behavior system from user IDs into resolved User objects.

### DIFF
--- a/packages/sdk/src/types/internal/actor.ts
+++ b/packages/sdk/src/types/internal/actor.ts
@@ -33,12 +33,11 @@ export class InternalActor implements InternalPatchable<ActorLike> {
         const behavior = (this.behavior && this.behavior.behaviorType === actionEvent.behaviorType)
             ? this.behavior : undefined;
         if (behavior && behavior._supportsAction(actionEvent.actionName)) {
-            behavior._performAction(actionEvent.actionName, actionEvent.actionState,
-                this.actor.context.user(actionEvent.userId));
+            behavior._performAction(actionEvent.actionName, actionEvent.actionState, actionEvent.user);
         } else {
             const action = (this.actor as any)[actionEvent.actionName.toLowerCase()] as DiscreteAction;
             if (action) {
-                action._setState(this.actor.context.user(actionEvent.userId), actionEvent.actionState);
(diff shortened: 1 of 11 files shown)


packages/sdk/src/types/internal/actor.ts
```diff
@@ -33,11 +33,12 @@ export class InternalActor implements InternalPatchable<ActorLike> {
         const behavior = (this.behavior && this.behavior.behaviorType === actionEvent.behaviorType)
             ? this.behavior : undefined;
         if (behavior && behavior._supportsAction(actionEvent.actionName)) {
-            behavior._performAction(actionEvent.actionName, actionEvent.actionState, actionEvent.userId);
+            behavior._performAction(actionEvent.actionName, actionEvent.actionState,
+                this.actor.context.user(actionEvent.userId));
```
tombuMS: This logic should just be moved to the `public 'recv-perform-action'` method in the execution protocol, and the `ActionEvent` interface should be updated to contain a user instead of a user id.

packages/sdk/src/types/internal/actor.ts
```diff
@@ -33,11 +33,12 @@ export class InternalActor implements InternalPatchable<ActorLike> {
...
             ? this.behavior : undefined;
         if (behavior && behavior._supportsAction(actionEvent.actionName)) {
-            behavior._performAction(actionEvent.actionName, actionEvent.actionState, actionEvent.userId);
+            behavior._performAction(actionEvent.actionName, actionEvent.actionState,
+                this.actor.context.user(actionEvent.userId));
         } else {
             const action = (this.actor as any)[actionEvent.actionName.toLowerCase()] as DiscreteAction;
             if (action) {
-                action._setState(actionEvent.userId, actionEvent.actionState);
+                action._setState(this.actor.context.user(actionEvent.userId), actionEvent.actionState);
```
tombuMS: Same as above.  Then it would just be `actionEvent.user` passed in here.

eanders-MS:
@stevenvergenz Samples need updating to reflect this change.

PR #293, opened 2019-04-05
[NOT FOR MERGE] User-exclusive actors design v1

stevenvergenz:


packages/sdk/src/types/runtime/actor.ts
```diff
@@ -80,20 +81,21 @@ export interface ActorSet {
...
 
-    private _emitter: events.EventEmitter;
+    private _emitter = new events.EventEmitter();
     /** @hidden */
     public get emitter() { return this._emitter; }
 
     private _name: string;
     private _tag: string;
-    private _parentId: string;
+    private _exclusiveToUser: User;
```
eanders-MS: >User [](start = 30, length = 4)

Let's store this as a string. It should be ok for actors to be assigned to users that aren't joined yet.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -80,20 +81,21 @@ export interface ActorSet {
...
+    private _parentId = ZeroGuid;
     private _subscriptions: SubscriptionType[] = [];
-    private _transform: ActorTransform;
-    private _appearance: Appearance;
+    private _transform = new ActorTransform();
+    private _appearance = new Appearance(this);
```
eanders-MS: >this [](start = 41, length = 4)

Is `this` accessible here? I thought it wasn't accessible until after super() was called in the ctor.

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: That's a good design idea, but user IDs are useless to users. I'd much rather give them a User object if I can. What if it's stored as an ID, then the getter resolves it to a User?

---
In reply to: [272676558](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/293#discussion_r272676558) [](ancestors = 272676558)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: I've never heard of that restriction, I don't know. It compiled, but I guess it could be null or something. I can check. I like initializing as much as possible in-line with the declaration though.

---
In reply to: [272676562](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/293#discussion_r272676562) [](ancestors = 272676562)

packages/sdk/src/types/runtime/actor.ts
eanders-MS: I think it needs to return a string so that it will function correctly when the user isn't in the space. The value returned by the getter should be consistent whether or not the user is in the space.

---
In reply to: [272693571](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/293#discussion_r272693571) [](ancestors = 272693571,272676558)

eanders-MS:
The "create-actors" sync stage may need additional logic to filter exclusive actors. There may be another complication: I don't think we have the userId yet while sync is happening.

stevenvergenz:
Oh damn, you're right. I guess exclusive actors shouldn't get synced in that phase, and we'll sync them after join.

---
In reply to: [480378144](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/293#issuecomment-480378144) [](ancestors = 480378144)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: I've changed this to a plain string. Union types are a pain to handle correctly, so at least a string can be handled consistently.

---
In reply to: [272696386](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/293#discussion_r272696386) [](ancestors = 272696386,272693571,272676558)

PR #303, opened 2019-04-12
[DO NOT MERGE] User-exclusive actors v2

stevenvergenz:
* Add the `exclusiveToUser` field on actors
* Add the `userExclusiveMessages` field on clients as a message queue
* Add `shouldSendToUser` to Rule objects, which indicates whether a given payload should be sent to a given user
* Add the protocol preprocessor `ClientDesyncPreprocessor` (added to ClientSync and ClientExecution), which queues up messages related to user-exclusive actors in the client, then when a user joins that client, it re-sends the messages targeting that user and throws out the rest.

TODO: Write per-payload `shouldSendToUser` filters

### DIFF
--- a/packages/sdk/src/adapters/multipeer/rules.ts
+++ b/packages/sdk/src/adapters/multipeer/rules.ts
@@ -77,6 +77,18 @@ export type Rule = {
          */
         beforeQueueMessageForClient: (
             session: Session, client: Client, message: any, promise: ExportedPromise) => Message;
+        /**
+         * Called twice before a message is sent: first to determine if a message is user-dependent
+         * (it is queued until user-join if so), and second to determine if the joined user is the
+         * correct user.
+         * @param message The message to be checked
+         * @param userId A GUID to a (possibly unjoined) user
+         * @param session The current session.
+         * @param client The client to send the message
+         * @returns `null` if the message does not depend on a user, `true` if it depends on the given
+         * user, and `false` if it depends on a different user.
+         */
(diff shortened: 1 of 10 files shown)


packages/sdk/src/types/runtime/actor.ts
```diff
@@ -80,20 +81,21 @@ export interface ActorSet {
  */
 export class Actor implements ActorLike, Patchable<ActorLike> {
     // tslint:disable:variable-name
-    private _internal: InternalActor;
+    private _internal = new InternalActor(this);
```
stevenvergenz: I tested the use of `this` in in-line initializers, and it all works as if it were initialized from inside the constructor.

packages/sdk/src/adapters/multipeer/rules.ts
```diff
@@ -1043,9 +989,6 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
                 client: Client,
                 message: Message<Payloads.UserJoined>
             ) => {
-                // Associate the client connection with the user id.
-                client.userId = message.payload.user.id;
```
eanders-MS: Why remove this?

stevenvergenz: It is not removed, so much as moved. This field is now populated earlier than this, in the clientDesyncPreprocessor.

---
In reply to: [275434399](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/303#discussion_r275434399) [](ancestors = 275434399)

PR #314, opened 2019-04-17
User-exclusive actors

stevenvergenz:
* Add an `exclusiveToUser` field to actors. If this field is given, messages pertaining to this actor will be filtered out for every client except for the one belonging to this user.
* Update functional tests with an inclusive/exclusive toggle. Inclusive is the same as before. In exclusive mode, all the test actors will be exclusive to the user that checked the toggle.


packages/sdk/src/adapters/multipeer/clientDesyncPreprocessor.ts
```diff
@@ -0,0 +1,47 @@
...
+        // this message is intended for a different user, discard
+    }
+
+    /** @hidden */
+    public beforeRecv(message: Message): Message {
+        if (message.payload.type === 'user-joined') {
+            const userJoin = message.payload as UserJoined;
+            this.client.userId = userJoin.user.id;
+            while (this.client.userExclusiveMessages.length > 0) {
+                const queuedMsg = this.client.userExclusiveMessages.splice(0, 1)[0];
```
eanders-MS: >this.client.userExclusiveMessages.splice(0, 1)[0] [](start = 34, length = 49)

This could be simplified to `this.client.userExclusiveMessages.splice(0);`

stevenvergenz: `splice(0)` would remove all elements from the array, and return a new array with those contents. The way this is structured, I just want to pop the first element from the array. I could structure it differently I suppose.

---
In reply to: [276690006](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/314#discussion_r276690006) [](ancestors = 276690006)

packages/sdk/src/adapters/multipeer/rules.ts
```diff
@@ -678,23 +635,17 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
                 client: Client,
                 message: Message<Payloads.ObjectSpawned>
             ) => {
-                // Check that this is the authoritative client.
-                if (client.authoritative) {
+                // Check that this is the authoritative client
+                const exclusiveUser = session.actorSet[message.payload.actors[0].id].exclusiveToUser;
```
eanders-MS: @stevenvergenz there's a regression here. `message.payload.actors` will be empty if the object failed to spawn (missing from library, glTF not found, etc.).

PR #319, opened 2019-04-24
Updated button event API for button vs click events

tombuMS:
This will allow for the differentiation of these button events for long press grabs and clicks on 3dof and 2d within host apps where button conflicts happen for grab and button interactions.

NOTE: The previous click event that was pressed and released is now a higher level click event that only happens when a long press has not happened and been handled by a grab.  A new button event has been added for raising events in the app on a button pressed and released.


tombuMS:
This is a fix for issue #[310](https://github.com/Microsoft/mixed-reality-extension-sdk/issues/310)

Issue #318, opened 2019-04-23
Subscriptions set as option doesn't work


norybiak:
```javascript
const actor = MRE.Actor.CreatePrimitive(this.context, 
{
    definition: 
    {

    },
    addCollider: true,
    actor: 
    {
        name: "Actor",
        
        //works
        subscriptions: [ 'transform' ]
    },

    //Does not work
    subscriptions: [ 'transform' ]
}).value;

//works
actor.subscribe('transform');
```

willneedit:
Suggestion: Remove subscriptions as an option outside of the actor definition. No need to have that kind of redundancy.

eanders-MS:
Good find, thanks.

PR #333, opened 2019-06-04
Make subscriptions only a part of the patch.

tombuMS:
This is a cleanup associated with some redundancy in the subscriptions API as well as moving the subscription updates in to the actor patch.

This is referenced in issue #[318](https://github.com/microsoft/mixed-reality-extension-sdk/issues/318)


stevenvergenz:
Should this PR also remove the UpdateSubscription payload type?

tombuMS:
> Shouldn't this PR also remove the UpdateSubscriptions payload type?

Yes it should.  Forgot to do that when I got back to this after being gone a month.  Good catch.

PR #334, opened 2019-06-04
add show and hide function

qulingyi:


packages/altspacevr-extras/src/videoPlayerManager.ts
```diff
@@ -41,11 +43,38 @@ export class VideoPlayerManager {
...
     private userJoined(user: User) {
-        if (Object.keys(this.videos).length > 0) {
-            const userRPC = new UserRPC(user);
-            for (const actorID in this.videos) {
-                if (this.videos.hasOwnProperty(actorID)) {
+        this.hasAnyUserJoined = true;
+        const userRPC = new UserRPC(user);
+        for (const actorID in this.videos) {
+            if (this.videos.hasOwnProperty(actorID)) {
+                if (this.state === "show") {
```
tombuMS: Any strings that do not need to be localized should be surrounded by single quotes.

tombuMS: Also the same of the other uses of double quotes below.

packages/altspacevr-extras/src/videoPlayerManager.ts
```diff
@@ -41,11 +43,38 @@ export class VideoPlayerManager {
...
+    //                     {
+    //                         parentId: actorID,
+    //                         URL: this.videos[actorID].url,
+    //                         startTime: this.videos[actorID].basisTime + Date.now() / 1000.0
+    //                     });
+    //             }
+    //         }
+    //     }
+    //     this.hasAnyUserJoined = true;
+    // }
```
tombuMS: Make sure to remove this commented out code before merging.

Issue #344, opened 2019-06-14
Expose kinematic on rigid body


eanders-MS:
Sometimes we don't want a rigid body to be simulated by the physics engine.

PR #346, opened 2019-06-15
Add in IsKinematic property to rigid body

tombuMS:
Referenced by issue #[344](https://github.com/microsoft/mixed-reality-extension-sdk/issues/344)

### DIFF
--- a/packages/sdk/src/types/runtime/rigidbody.ts
+++ b/packages/sdk/src/types/runtime/rigidbody.ts
@@ -35,7 +35,7 @@ export interface RigidBodyLike {
     useGravity: boolean;
     /** 
      * Whether the rigid body is kinematic.  Note kinematic rigid bodies participate in collisions,
-     * but are not simulated by the rigid body.  This is usefull for objects that should collide with
+     * but are not simulated by the rigid body.  This is useful for objects that should collide with
      * other objects, but you want to control the position/rotation manually or animate them.
      */
     isKinematic: boolean;
@@ -96,7 +96,7 @@ export class RigidBody implements RigidBodyLike {
      */
 
     /**
-     * INTERNAL USE ONLY.
+     * @hidden
      * Creates a new RigidBody instance.
      * @param $owner The owning actor instance. Field name is prefixed with a dollar sign so that it is ignored by
      * the actor patch detection system.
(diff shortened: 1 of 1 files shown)


packages/sdk/src/types/runtime/rigidbody.ts
```diff
@@ -23,6 +23,7 @@ export interface RigidBodyLike {
     detectCollisions: boolean;
     collisionDetectionMode: CollisionDetectionMode;
     useGravity: boolean;
+    isKinematic: boolean;
```
stevenvergenz: This whole interface needs documentation, but could you start here? Use Material for reference: [https://github.com/microsoft/mixed-reality-extension-sdk/blob/red/packages/sdk/src/types/runtime/assets/material.ts](https://github.com/microsoft/mixed-reality-extension-sdk/blob/red/packages/sdk/src/types/runtime/assets/material.ts)

tombuMS: Sure.  I am fine with beginning the documentation updates.

packages/sdk/src/types/runtime/rigidbody.ts
```diff
@@ -15,38 +15,75 @@ import {
...
     mass: number;
+    /** Whether to detect collisions with this rigid body. */
     detectCollisions: boolean;
+    /** The collision detection mode to use with this rigid body.  @see CollisionDetectionMode for options. */
     collisionDetectionMode: CollisionDetectionMode;
+    /** Whether the rigid body is affected by gravity. */
     useGravity: boolean;
+    /** 
+     * Whether the rigid body is kinematic.  Note kinematic rigid bodies participate in collisions,
+     * but are not simulated by the rigid body.  This is usefull for objects that should collide with
```
eanders-MS: spelling: usefull -> useful

packages/sdk/src/types/runtime/rigidbody.ts
```diff
@@ -59,6 +96,7 @@ export class RigidBody implements RigidBodyLike {
      */
 
     /**
+     * INTERNAL USE ONLY.
```
eanders-MS: Use @hidden instead to hide from doc generation

packages/sdk/src/types/runtime/rigidbody.ts
eanders-MS: Thanks for adding documentation.

packages/sdk/src/types/runtime/rigidbody.ts
tombuMS: Ok.  I thought that was what should be done, but I was following the example of the material.ts file like Steven had said.  I will just replace the internal use with @hidden then.

Issue #320, opened 2019-04-25
Grab: grabbed object transform changes should lerp on other clients


eanders-MS:
When an object is grabbed and manipulated, transform changes are synced by the client doing the grabbing up to the app and then down to the other clients. When the transform change is applied remotely, it is overwritten rather than interpolated, causing a visible jump. It should be interpolated so that the object moves smoothly.

> Side Bar: An even better solution would be is grab worked like attachments, where no transform sync is needed, but that requires a change to the way Altspace implements grab. Either way, Altspace's implementation details are leaking through here.

PR #350, opened 2019-06-17
Grab lerping on synchronized clients

tombuMS:
This adds in lerping for the synchronized clients during a grab.  This is performed through the existing actor correction payload that is no fully hooked up and supported.

Referenced by issue #[320](https://github.com/microsoft/mixed-reality-extension-sdk/issues/320)

### DIFF
--- a/packages/sdk/src/adapters/multipeer/rules.ts
+++ b/packages/sdk/src/adapters/multipeer/rules.ts
@@ -279,6 +279,9 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
                             }
                         }
                     });
+
+                    // We have merged the actor correction in to an existing actor update.  We do not want to
+                    // propagate the correction message further.
                     return undefined;
                 }
                 return message;
@@ -357,6 +360,9 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
                 if (queuedMessage) {
                     const existingPayload = queuedMessage.message.payload as Partial<Payloads.ActorUpdate>;
                     existingPayload.actor = deepmerge(existingPayload.actor, payload.actor);
+
+                    // We have merged the actor update in to an existing actor update.  We do not want to
+                    // propagate the update message further.
                     return undefined;
                 }
                 return message;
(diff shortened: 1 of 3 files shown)


packages/sdk/src/adapters/multipeer/rules.ts
```diff
@@ -257,35 +257,75 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
...
+                    existingPayload.actor = deepmerge(existingPayload.actor,  {
+                        payload: {
+                            actor: {
+                                transform: {
+                                    app: message.payload.appTransform
+                                }
+                            }
+                        }
+                    });
+                    return undefined;
```
eanders-MS: >return undefined; [](start = 20, length = 17)

Can you document why you're returning undefined here?

tombuMS: I got this from the actor-update so I will update with a comment there as well.

---
In reply to: [294896385](https://github.com/microsoft/mixed-reality-extension-sdk/pull/350#discussion_r294896385) [](ancestors = 294896385)

packages/sdk/src/adapters/multipeer/rules.ts
```diff
@@ -257,35 +257,75 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
...
             ) => {
-                message.payload.type = 'actor-update';
-                client.queueMessage(message, promise);
-                return undefined;
+                // Coalesce this actor correction with the previously queued update if it exists, maintaining a single
+                // update for this actor rather than queuing a series of them.  This is fine, as we do not need to lerp
+                // an actor correction on a late join user.  It can just be the updated actor values.
+                const payload = message.payload;
+                const queuedMessage = client.queuedMessages
+                    .filter(value =>
```
stevenvergenz: >filter [](start = 21, length = 6)

Why not a simple `find` instead of `filter` and `shift`?

tombuMS: I simply replicated what was already wrote for the actor-update rule.  @eanders-MS would be a better person to answer his original reasoning in doing it this way.

eanders-MS: An oversight on my part for not commenting it initially. Thanks for correcting it.

PR #361, opened 2019-06-28
Maintain app/local transform sync in multipeer session

stevenvergenz:
* When it comes to glTFs and other implicitly created actors, don't overwrite the pristine load state with the patch's placeholders for the same state. This was assigning default materials and overwriting the client-only materials from the glTF.
* When caching transform updates in the session, only save one of app and local coordinates. Saving both requires recomputing after asymmetric updates, which is unnecessarily complicated.

### DIFF
--- a/packages/sdk/src/adapters/multipeer/session.ts
+++ b/packages/sdk/src/adapters/multipeer/session.ts
@@ -257,6 +257,16 @@ export class Session extends EventEmitter {
             // Merge the update into the existing actor.
             syncActor.initialization.message.payload.actor
                 = deepmerge(syncActor.initialization.message.payload.actor, message.payload.actor);
+
+            // strip out transform data that wasn't updated
+            // so it doesn't desync from the updated one
+            const cacheTransform = syncActor.initialization.message.payload.actor.transform;
+            const patchTransform = message.payload.actor.transform;
+            if (patchTransform && !patchTransform.local) {
+                cacheTransform.local = undefined;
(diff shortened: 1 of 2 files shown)


packages/sdk/src/adapters/multipeer/session.ts
```diff
@@ -257,6 +257,16 @@ export class Session extends EventEmitter {
             // Merge the update into the existing actor.
             syncActor.initialization.message.payload.actor
                 = deepmerge(syncActor.initialization.message.payload.actor, message.payload.actor);
+
+            // strip out transform data that wasn't updated
+            // so it doesn't desync from the updated one
+            const cacheTransform = syncActor.initialization.message.payload.actor.transform;
```
eanders-MS: >cacheTransform [](start = 18, length = 14)

I believe it's possible for this to be undefined, depending on how the syncActor was created.

stevenvergenz: It will only be undefined if patchTransform is also undefined, because it's after the deepmerge.

---
In reply to: [299102576](https://github.com/microsoft/mixed-reality-extension-sdk/pull/361#discussion_r299102576) [](ancestors = 299102576)

eanders-MS: That's true. Thanks.

---
In reply to: [299124936](https://github.com/microsoft/mixed-reality-extension-sdk/pull/361#discussion_r299124936) [](ancestors = 299124936,299102576)

PR #369, opened 2019-07-03
[DO NOT MERGE] Assets v2 proposal

stevenvergenz:
My proposal for an asset system infrastructure overhaul, which includes the ability to unload assets.

### DIFF
--- a/packages/sdk/src/types/runtime/assets/assets-v2.md
+++ b/packages/sdk/src/types/runtime/assets/assets-v2.md
@@ -1,0 +1,95 @@
+# Asset Unloading
+
+This proposal details the API changes required to allow developers to explicitly request
+that a given asset (material, texture, sound) be unloaded from memory on the clients,
+and unassign all references to them.
+
+Today, assets can be created individually, or created en masse by loading a container
+(glTF) for assets. This proposal does not have separate unload APIs for these two cases.
+Instead, during synchronization, the adapter will examine which assets are loaded at sync
+time, send a container load request if at least one asset from that container is still
+live, then send unloads for all the unused ones. This is processor-inefficient, but
+maintains API focus on the concept of assets.
+
+An alternative to the above would be to modify `AssetManager.loadGltf` to allow the
+developer to specify which assets are loaded. This is technically possible and solves
+the inefficiencies of the above, but requires developer knowledge of glTF internals,
+which may not be feasible. As another alternative, the API could require that
+container-loaded assets be unloaded as a block. This alternative doesn't really add
+any new functionality to the proposal, but reduces multipeer adapter complexity.
+
+## Proposed API Changes
+
+1. `AssetManager.unloadAssets(...assetIds: Asset[]): Promise<void>`
+
+	This method does four things:
+
+	1. It breaks all references to each asset by using the asset's internal reference
+		tracker, which will store references to each actor/asset that uses the asset.
+	2. It waits for the next batch of updates to propagate to the clients, so the assets
+		to be unloaded are not in use during the unload. This wait is done via a new
+		context method that resolves a promise when the next update is done.
+	3. It sends an `unload-assets` message to clients with the asset IDs. This will
+		trigger an `Object.Destroy` in Unity's asset cache, freeing the memory.
+	4. It collects success/failure messages from the client, and propagates any errors
+		from the client unload.
+
+2. `AssetManager.unloadGroup(group: AssetGroup): Promise<void>`
+
+	This simply calls `unloadAssets` with the IDs of all the assets in the group. If
+	alternative two is implemented, this will be the only way to unload container-loaded
+	assets. Even if it isn't, this will be more efficient than piecemeal unloading, but
+	the developer can decide which is more appropriate.
+
+3. `Asset.unload(): Promise<void>`
+
+	Calls `unloadAssets` with the current asset ID.
+
+## Proposed Sync Layer Changes
+
+### Session
+
+Currently the session stores two things related to assets: a list of asset creation
+messages (`load-assets` and `create-asset` payloads), and a map of asset updates
+since creation by ID. This proposal would replace the creation list with a creation
+map indexed by message ID, and would replace the update map with a map of new
+SyncAsset structures. These would contain an asset ID, a creation message ID, a flag
+indicating if the asset has been uploaded, and a cumulative update message. Session
+state would get modified through four functions:
+
+1. `cacheAssetCreationRequest(message: Message<LoadAssets | CreateAsset>): void`
+
+	Store the creation message in the session by message ID.
+
+2. `cacheAssetCreation(assetId: string, creationMessageId: string): void`
+
+	For each asset resulting from a creation message, associate the asset with the
+	creation message in the session.
+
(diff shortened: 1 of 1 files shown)


packages/sdk/src/types/runtime/assets/assets-v2.md
```diff
@@ -0,0 +1,100 @@
...
+
+This proposal details the API changes required to allow developers to explicitly request
+that a given asset (material, texture, sound) be unloaded from memory on the clients,
+and unassign all references to them.
+
+Today, assets can be created individually, or created en masse by loading a container
+(glTF) for assets. This proposal does not have separate unload APIs for these two cases.
+Instead, during synchronization, the adapter will examine which assets are loaded at sync
+time, send a container load request if at least one asset from that container is still
+live, then send unloads for all the unused ones. This is processor-inefficient, but
```
sorenhan: "unload the unused ones" doesn't make sense to me. I would expect that assets loaded on machine A are then also loaded on machine B during synchronization. Otherwise, what if dev had preloaded an asset but aren't using it right at this moment? It should still be ready to instantly instantiate on machine B

stevenvergenz: That is a phrasing error. It is not *unused* assets that are unloaded here. It's *unloaded*, i.e. assets that previously had `unload` called on them.

packages/sdk/src/types/runtime/assets/assets-v2.md
```diff
@@ -0,0 +1,100 @@
...
+
+	1. It breaks all references to each asset by using the asset's internal reference
+		tracker, which will store references to each actor/asset that uses the asset.
+	2. It waits for the next batch of updates to propagate to the clients, so the assets
+		to be unloaded are not in use during the unload. This wait is done via a new
+		context method that resolves a promise when the next update is done.
+	3. It sends an `unload-assets` message to clients with the asset IDs. This will
+		trigger an `Object.Destroy` in Unity's asset cache, freeing the memory.
+	4. It collects success/failure messages from the client, and propagates any errors
+		from the client unload.
```
sorenhan: I think UnloadAssets() is a confusing API, if Asset[] can either be top-level (a preloaded GLTF, or a programmatically created), or owned by a higher level asset (the GLTF owns textures, materials etc). 
I think we should either use unloadGroup() only, or consider having individual assets are all creating a TopLevelAsset or AssetContainer or something like that, and the only thing you can free are asset containers. I think unloadGroup() is simpler

packages/sdk/src/types/runtime/assets/assets-v2.md
```diff
@@ -0,0 +1,100 @@
...
+	alternative two is implemented, this will be the only way to unload container-loaded
+	assets. Even if it isn't, this will be more efficient than piecemeal unloading, but
+	the developer can decide which is more appropriate.
+
+3. `AssetManager.unloadUnused(): Promise<void>`
+
+	Loops over all assets, finds those that have no references, and calls `unloadAssets`
+	with their IDs.
+
+4. `Asset.unload(): Promise<void>`
```
sorenhan: >: Promise<void> [](start = 18, length = 15)

Do we actually need to make these return Promise<void>? are there any cases you may want to wait on this?

packages/sdk/src/types/runtime/assets/assets-v2.md
```diff
@@ -0,0 +1,100 @@
...
+	`session.cacheAssetCreation`.
+3. On outgoing `asset-update` messages, call `session.cacheAssetUpdate`.
+4. On outgoing `unload-assets` messages, call `session.cacheAssetUnload`.
+
+### Client Sync
+
+During the `load-assets` stage of client sync:
+
+1. Send creation messages of all loaded assets.
+2. Send unload messages for all container-loaded assets that have been unloaded.
```
sorenhan: Does this mean that if you load and unload an asset 100 of times, it will send 100 load and 100 unload requests? Instead, I'd expect the sync layer to remove the cacheassetcreationrequest/cacheassetcreation/cacheassetupdate for every asset in that assetcontainer/group.

packages/sdk/src/types/runtime/assets/assets-v2.md
stevenvergenz: I understand and discuss the performance argument, but you think the API design is confusing? From the developer perspective, the only objects that are useful to building an experience are at the asset level. Making them track groups as well seems unnecessarily burdensome to me, given that there's no technical requirement for it.

---
In reply to: [300189061](https://github.com/microsoft/mixed-reality-extension-sdk/pull/369#discussion_r300189061) [](ancestors = 300189061)

packages/sdk/src/types/runtime/assets/assets-v2.md
stevenvergenz: Hmm, probably not. The only use case that comes to mind is if you want to wait until after unload to load in new stuff.

---
In reply to: [300189356](https://github.com/microsoft/mixed-reality-extension-sdk/pull/369#discussion_r300189356) [](ancestors = 300189356)

packages/sdk/src/types/runtime/assets/assets-v2.md
stevenvergenz: Loading a glTF multiple times is not a use case that is currently supported (I hope to support it with this update), but I would expect each load request to produce a duplicate set of assets, not the same set over and over. But to the spirit of your question, if a container contains 100 assets and 99 are unloaded, the sync layer will trigger the load of all 100, then immediately unload 99. If all 100 are unloaded, none get loaded because the whole container got flushed from the cache. So it works as you expect.

---
In reply to: [300189903](https://github.com/microsoft/mixed-reality-extension-sdk/pull/369#discussion_r300189903) [](ancestors = 300189903)

packages/sdk/src/types/runtime/assets/assets-v2.md
stevenvergenz: That is a phrasing error. It is not unused assets that are unloaded here. It's unloaded, i.e. assets that previously had `unload` called on them.

---
In reply to: [300188310](https://github.com/microsoft/mixed-reality-extension-sdk/pull/369#discussion_r300188310) [](ancestors = 300188310)

packages/sdk/src/types/runtime/assets/assets-v2.md
eanders-MS: The design of UnloadAssets goes contrary to the way content managers typically work in game engines. Loads and unloads should be symmetric. The complexity of the API makes it less discoverable, and also a burden on future runtime implementors (not all content loaders will support partial unloads). I'm in favor of the asset container approach, and only supporting unload at the container level.

packages/sdk/src/types/runtime/assets/assets-v2.md
eanders-MS: If unload is only supported at the container level, then this complexity goes away, correct?

stevenvergenz: This would be a no-op, yes.

---
In reply to: [300767341](https://github.com/microsoft/mixed-reality-extension-sdk/pull/369#discussion_r300767341) [](ancestors = 300767341)

PR #372, opened 2019-07-06
Video player Refactor

sorenhan:
Implemented a cleaner video player interface to replace the current AltspaceVR-exclusive RPC-based one. This is now complete.  This is a breaking change, in that old MREs would not play sounds (as payload has changed and been renamed)


eanders-MS:
> export class Sound extends Asset implements SoundLike, Patchable<AssetLike> {

Do you plan to rename this class too?

---
Refers to: packages/sdk/src/types/runtime/assets/sound.ts:16 in d139a62. [](commit_id = d139a620ec6734a7c18d8fb4a154cd33a051d71d, deletion_comment = False)

packages/sdk/src/types/runtime/assets/videoStream.ts
```diff
@@ -0,0 +1,106 @@
...
+import { Patchable } from '../../patchable';
+
+export interface VideoStreamLike {
+	videoSourceType: VideoSourceType;
+	uri: string;
+	duration: number;
+}
+
+/** Type describing how to interpret the video Source URI. */
+export enum VideoSourceType {
```
eanders-MS: >VideoSourceType [](start = 12, length = 15)

why do we need to know the video type?

eanders-MS:
> export class Sound extends Asset implements SoundLike, Patchable<AssetLike> {

Sorry wrong file. Deleting comment.

---
In reply to: [509274895](https://github.com/microsoft/mixed-reality-extension-sdk/pull/372#issuecomment-509274895) [](ancestors = 509274895)

---
Refers to: packages/sdk/src/types/runtime/assets/sound.ts:16 in d139a62. [](commit_id = d139a620ec6734a7c18d8fb4a154cd33a051d71d, deletion_comment = False)

packages/sdk/src/sound/setMediaStateOptions.ts
```diff
@@ -51,4 +52,15 @@ export type SetSoundStateOptions = {
 	 */
 	rolloffStartDistance?: number;
 
+	/**
+	 * The media should start at, or seek this many seconds into the media.
+	 * Time is in seconds relative to start of clip.
+	 */
+		Time?: number;
```
eanders-MS: >Time [](start = 2, length = 4)

lowercase: 'time'

packages/sdk/src/sound/setMediaStateOptions.ts
```diff
@@ -51,4 +52,15 @@ export type SetSoundStateOptions = {
...
+
+	/**
+	 * Should the video stream be visible or invisible
+	 * Does not apply to sounds, only to video streams
+	 */
+	Visible?: boolean;
```
eanders-MS: >Visible [](start = 1, length = 7)

lowercase; 'visible'

packages/sdk/src/sound/setMediaStateOptions.ts
```diff
@@ -4,9 +4,9 @@
  */
 
 /**
- * Parameters to the `Sound.setSoundState` and `Actor.startSound` calls.
+ * Parameters to the `MediaInstance.setState`, `Actor.startSound`, and `Actor.startVideoStream` calls.
  */
-export type SetSoundStateOptions = {
+export type SetMediaStateOptions = {
```
eanders-MS: >SetMediaStateOptions [](start = 12, length = 20)

SetMediaStateOptions is becoming a confusion of different types. I'd like to see this factored out into base a base type and media-type-specific types:
```
export type SetMediaStateOptions = {
   // ...common fields
};

export type SetAudioStreamStateOptions = SetMediaStateOptions & {
   // ...audio-specific options
   doppler: number;
   pitch: number;
};

export type SetVideoStreamStateOptions = SetMediaStateOptions & {
   // ...video-specific options
};

File: packages/sdk/src/types/runtime/mediaInstance.ts [hunk 5]
```diff
@@ -5,11 +5,11 @@
 
 import UUID from 'uuid/v4';
 import { log } from '../../log';
-import { SetSoundStateOptions, SoundCommand } from '../../sound';
+import { MediaCommand, SetMediaStateOptions } from '../../sound';
 import { createForwardPromise, ForwardPromise } from '../forwardPromise';
 import { Actor } from './actor';
 
-export class SoundInstance {
+export class MediaInstance {
```
eanders-MS: >MediaInstance [](start = 13, length = 13)

Class needs documentation

packages/sdk/src/types/runtime/assets/videoStream.ts
sorenhan: Because the URI is not always a raw URL, and because it's clearer that we are doing special solution based on the type. For VideoSourceType=YouTube you can use URI `1nCqRmx3Dnw`, rather than `[https://www.youtube.com/watch?v=1nCqRmx3Dnw`](https://www.youtube.com/watch?v=1nCqRmx3Dnw`). And the type indicates that the player applies custom logic for choosing ideal stream for each user rather than one-bandwidth-fits-all


---
In reply to: [301160693](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/372#discussion_r301160693) [](ancestors = 301160693)

packages/sdk/src/sound/setMediaStateOptions.ts
sorenhan: If there were 3 or more media types, I could see an argument for splitting it up. Considering there are only 2 types, and there are only 1 audio-specific option (Doppler) and 2 videostream-specific options (Visible and Looping), I think that the code complexity will be higher than the benefit.

---
In reply to: [301167926](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/372#discussion_r301167926) [](ancestors = 301167926)

eanders-MS: Animation will end up using this, correct? That makes three. Users will try to apply doppler to animations.

---
In reply to: [301348602](https://github.com/microsoft/mixed-reality-extension-sdk/pull/372#discussion_r301348602) [](ancestors = 301348602,301167926)

packages/sdk/src/types/runtime/assets/videoStream.ts
eanders-MS: I don't think we need to know or care about this on the SDK side.

---
In reply to: [301347617](https://github.com/microsoft/mixed-reality-extension-sdk/pull/372#discussion_r301347617) [](ancestors = 301347617,301160693)

eanders-MS: Seems like we're leaking an Altspace implementation detail.

---
In reply to: [301350379](https://github.com/microsoft/mixed-reality-extension-sdk/pull/372#discussion_r301350379) [](ancestors = 301350379,301347617,301160693)

eanders-MS: I don't think we should expose this. The client can encode it in the url.

---
In reply to: [301350449](https://github.com/microsoft/mixed-reality-extension-sdk/pull/372#discussion_r301350449) [](ancestors = 301350449,301350379,301347617,301160693)

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -648,7 +662,7 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 		if (from.collider) this._setCollider(from.collider);
 		if (from.text) this.enableText(from.text);
 		if (from.lookAt) this.enableLookAt(from.lookAt.actorId, from.lookAt.mode);
-		if (from.grabbable) this._grabbable = from.grabbable;
+		if (from.grabbable !== undefined) this._grabbable = from.grabbable;
```
stevenvergenz: Good catch!

packages/sdk/src/types/runtime/assets/videoStream.ts
```diff
@@ -0,0 +1,81 @@
...
+	private _internal = new InternalAsset(this);
+	// tslint:enable:variable-name
+
+	/** @hidden */
+	public get internal() { return this._internal; }
+
+	/** The URI, if any, this videostream was loaded from */
+	public get uri() { return this._uri; }
+
+	/** The length the loaded videostream */
```
stevenvergenz: What units, milliseconds or seconds?

packages/sdk/src/types/network/payloads/payloads.ts
```diff
@@ -359,13 +358,13 @@ export type SyncAnimations = Payload & {
  * @hidden
  * App to engine. Starts playing a sound.
  */
-export type SetSoundState = Payload & {
-	type: 'set-sound-state';
+export type SetMediaState = Payload & {
+	type: 'set-media-state';
 	id: string;
 	actorId: string;
 	soundAssetId: string;
```
stevenvergenz: Rename this field to be generic. `MediaAssetId`?

stevenvergenz:
We'll have to coordinate this merge with my asset unload changes. Both these PRs are probably going out in the same release.

PR #383, opened 2019-07-10
Suggested video player tweaks

eanders-MS:
Will be easier to review if I open a PR.


packages/functional-tests/src/tests/video-test.ts
```diff
@@ -33,16 +33,14 @@ export default class VideoTest extends Test {
 		const videoStreamPromise1 = this.app.context.assetManager.createVideoStream(
 			'group1',
 			{
-				videoSourceType: MRE.VideoSourceType.YouTube,
-				uri: `1roy4o4tqQM`
-				// `z1YNh9BQVRg`
+				uri: `youtube://1roy4o4tqQM`
+				// `youtube://z1YNh9BQVRg`
```
sorenhan: i liked how you got rid of the unnecessary videosourcetype variable - thumbs up. We still need to list explicitly which protocols need to be supported to be "MRE VideoPlayerFactory compliant". Hopefully, at some point, we can find time to move the Youtube/Mixer/Twitch URI->Raw stream lookup logic over to TypeScript, and the only needed protocol will be raw streams.

eanders-MS: Agreed that would be a nice refinement.

PR #384, opened 2019-07-11
Asset Unloading

stevenvergenz:
* Merged `AssetManager`, `AssetGroup`, and `AssetCollection` into a single class: `AssetContainer`.
* Made `AssetContainer` not a singleton. To create assets, create an `AssetContainer` first, then call its create/load functions.
* Added asset reference counting, and the ability for each asset type to dereference itself from any actors/assets that use it.
* Added an `unload` method to `AssetContainer` that de-references, then triggers the unload of, all assets in the container.
* Add dummy optional fields to the `Asset` base class.
* Add `nextUpdate` method to `InternalContext`, which returns a promise that resolves after the next `InternalContext.update` call.

**NOTE**: This change is compatible with MRE-Unity 0.11. It only adds new fields and new messages, it does not change anything the current system uses.


packages/sdk/src/types/runtime/assets/asset.ts
```diff
@@ -82,12 +84,51 @@ export abstract class Asset implements AssetLike {
...
+	public get mesh(): Mesh { return null; }
+	/** @inheritdoc */
+	public get material(): Material { return null; }
+	/** @inheritdoc */
+	public get texture(): Texture { return null; }
+	/** @inheritdoc */
+	public get sound(): Sound { return null; }
+
+	/** A promise that resolves when the asset is finished loading */
+	public get created() { return this._loadedPromise; }
```
eanders-MS: >created() [](start = 12, length = 9)

Model this after the way it's done in Actor. That implementation handles the load failure case, where future calls to created() will reject if the load failed.

eanders-MS: Oh wait, your's does too, just a different way. Deleting comment.

---
In reply to: [302370180](https://github.com/microsoft/mixed-reality-extension-sdk/pull/384#discussion_r302370180) [](ancestors = 302370180)

eanders-MS:
Add an asset unload test?

stevenvergenz:
Those tests that use assets I've refactored to unload those assets on test completion.

---
In reply to: [510360473](https://github.com/microsoft/mixed-reality-extension-sdk/pull/384#issuecomment-510360473) [](ancestors = 510360473)

packages/sdk/src/types/runtime/assets/assetContainer.ts
```diff
@@ -0,0 +1,198 @@
...
+
+	/** Break references to all assets in the container, and unload them to free memory */
+	public unload(): void {
+		for (const a of this.assets) {
+			a.breakAllReferences();
+		}
+		this.context.internal.assetContainers.delete(this);
+		this._assets = null;
+
+		this.context.internal.nextUpdate().then(() => {
```
eanders-MS: >this.context.internal.nextUpdate().then(() => { [](start = 2, length = 47)

Please document why you need to wait until after the next update.

packages/sdk/src/types/network/payloads/assets.ts
```diff
@@ -9,16 +9,26 @@ import { AssetLike, AssetSource } from '../../runtime/assets';
 
 export type CreateColliderType = ColliderType | 'none';
 
+export type AssetPayloadType
```
eanders-MS: >AssetPayloadType [](start = 12, length = 16)

Mark @hidden

packages/sdk/src/types/runtime/assets/assetContainer.ts
```diff
@@ -0,0 +1,198 @@
...
+
+	/**
+	 * Load the assets in a glTF file by URL, and this container with the result.
+	 * @param uri The URI to a glTF model.
+	 * @param colliderType The shape of the generated prefab collider.
+	 * @returns A promise that resolves with the list of loaded assets.
+	 */
+	public async loadGltf(uri: string, colliderType?: Payloads.CreateColliderType): Promise<Asset[]> {
+		if (!this._assets) {
+			throw new Error("Cannot load new assets into an unloaded container!");
```
eanders-MS: >Cannot load new assets into an unloaded container! [](start = 20, length = 50)

Why not? Seems like we should support this.

eanders-MS: This limitation seems artificial. What would it take to remove it?

---
In reply to: [303661613](https://github.com/microsoft/mixed-reality-extension-sdk/pull/384#discussion_r303661613) [](ancestors = 303661613)

stevenvergenz: Because semantically you're unloading the container, not assets. Shouldn't the container itself be destroyed afterward? You're right that it's artificial, but it feels right in context. Besides, it might prevent some weird timing things if somehow the create gets processed before the unload.

---
In reply to: [303665679](https://github.com/microsoft/mixed-reality-extension-sdk/pull/384#discussion_r303665679) [](ancestors = 303665679,303661613)

eanders-MS: We should think of it as unloading the contents of the container, not disposing the container.

---
In reply to: [303668263](https://github.com/microsoft/mixed-reality-extension-sdk/pull/384#discussion_r303668263) [](ancestors = 303668263,303665679,303661613)

eanders-MS: Supporting this would allow for asset container reuse, which is a nice convenience. I think we should allow it.

---
In reply to: [303675156](https://github.com/microsoft/mixed-reality-extension-sdk/pull/384#discussion_r303675156) [](ancestors = 303675156,303668263,303665679,303661613)

eanders-MS: Withdrawing my objection. The asset container needs to be cleaned up, and `unload` is the best place to do it.

---
In reply to: [303677566](https://github.com/microsoft/mixed-reality-extension-sdk/pull/384#discussion_r303677566) [](ancestors = 303677566,303675156,303668263,303665679,303661613)

Issue #162, opened 2019-02-01
Improve animation synchronization


eanders-MS:
The way animation timing is synchronized right now is subject to some error due to network latencies. It also has a weak point where it relies on a peer's response, which could fail. This should be improved.

sorenhan:
This should be redone to match the soundinstance synchronization pattern

sorenhan:
#239 should be completed before this work is completed, as animation needs the same abilities as sound.

PR #387, opened 2019-07-17
[RFC] Animation system - proposed changes

eanders-MS:
Howdy team. I want to make our animation system more useful. This design will affect how the approved task #162 will be handled, but goes beyond it by replacing our existing `createAnimation` and `animateTo` calls with a new method for interpolating any mutable property on `Actor` or `Asset`.

Please read when you have a chance and feedback.

Once this design is settled, some of it will be implemented with #162, but probably not all of it.


animation-proposal.md
```diff
@@ -0,0 +1,338 @@
...
+
+// Play the 'victory' animation, and wait for it to complete.
+await actor.animation("victory").start();
+```
+
+### Interpolations
+
+#### Add `Context.interpolate` method
+
+Add an `interpolate` method to class `Context`. Calling `interpolate` initiates a lerp operation on the host and returns a Promise that will resolve after `seconds` time has elapsed or the interpolation is otherwise canceled.
```
stevenvergenz: So the animation is happening server-side, lerped with a similar mechanism to `actor-correction` today?

animation-proposal.md
```diff
@@ -0,0 +1,338 @@
...
+
+```ts
+/**
+ * Interpolates the field indicated by `selector` to `value` over `interval` seconds along `options.curve`.
+ * @param selector The selector of the field to interpolate. Get this value by calling actor.selector(fieldpath).
+ * @param value The value to interpolate to or by, depending on the value of options.basis. Can be a selector or a concrete value.
```
stevenvergenz: Recommend naming this `to`, and adding an optional `from` argument, in case you want to reset the value and then animate it.

animation-proposal.md
```diff
@@ -0,0 +1,338 @@
...
+ * @param interval The duration of the interolation in seconds. Pass in `null` to cancel an existing matching interpolation.
+ * @param options Optional parameters to further customize the interpolation.
+ * @param options.curve Specifies the interpolation's speed curve. This can be one of the predefined curve names, or a custom curve. Default: "linear".
+ * @param options.basis How to interpret `value`. If `options.basis` is "absolute" this function will interpolate to `value`. If `options.basis` is "relative", this function interpolates by `value` amount. Default: "absolute".
+ * @param options.iterationCount How many times to run the interpolation. Can be a number, or "infinite". Default: 1.
```
stevenvergenz: js/ts has an `Infinity` value, recommend using this over a string.

animation-proposal.md
```diff
@@ -0,0 +1,338 @@
...
+ * @param options.direction The direction the interpolation should be played.
```
stevenvergenz: Can't animate anywhere but forward without both a from and a to value.

animation-proposal.md
```diff
@@ -0,0 +1,338 @@
...
+```ts
+/**
+ * A union of the different types that can be passed to the
+ * `interpolate` method's `value` parameter.
+ */
+export type InterpolationValueType
+	/** A string type indicates the value is a field selector. */
+	= string
+	/** The rest represent concrete final values. */
+	| Quaternion
```
stevenvergenz: Quaternions would be spherically interpolated I presume?

animation-proposal.md
```diff
@@ -0,0 +1,338 @@
...
+
+/**
+ * Describes how the `value` parameter to `interpolate` should be interpreted.
+ */
+export type InterpolationBasis
+	/** The value is an absolute value. The function operates as "Interpolate to this value". */
+	= "absolute"
+
+	/** The value is relative to the starting value. The function operates as "Interpolate by this much". */
+	| "relative"
```
stevenvergenz: Does this apply to quaternions?

animation-proposal.md
```diff
@@ -0,0 +1,338 @@
...
+// Cancel an interpolation. This applies to all types of interpolations, but
+// is particularly useful for stopping continuous interpolations.
+this.context.interpolate(
+	actor.selector("transform.app.position"),
+	target.selector("transform.app.position"),
+	null);
+)
+```
+
+#### Add `Actor.selector` and `Asset.selector` methods
```
stevenvergenz: This is basically the same function on both classes, right? Could it be a utility method instead of putting it in the prototype?

animation-proposal.md
```diff
@@ -0,0 +1,338 @@
...
+// Play the 'jump' animation, then do something when it's done.
+actor.animation("jump").start().then(() => console.log("done jumping"));
+
+// Play the 'victory' animation, and wait for it to complete.
+await actor.animation("victory").start();
+```
+
+### Interpolations
+
+#### Add `Context.interpolate` method
```
stevenvergenz: Context seems like an awkward place to put this. Util method? Or maybe a static method on the Animation class?

animation-proposal.md
eanders-MS: Animation/interpolation is happening client side.

animation-proposal.md
eanders-MS: That seems reasonable. This value needs to be sent to C#. I wonder how it will be expressed on the other side.

animation-proposal.md
eanders-MS: Correct. I mention this in the Examples section in a code comment.

animation-proposal.md
eanders-MS: Quaternions are a weird one. I want to support relative rotations. I wonder if we need to support passing a set of Euler angles as the `to` value.

animation-proposal.md
eanders-MS: They do the same thing, but the root of the selector string they generate is different, and they may need to access private info. A selector generated by an Actor will look something like: "actor.<actorId>.transform.app.position", whereas a Material might generate:
"material.<materialId>.<containerId>.mainTextureOffset.u".

animation-proposal.md
eanders-MS: Agreed. I mention below the idea of adding it to both Actor and Asset, to make it more discoverable. Note that interpolations are no longer related to animations in this new system, so it wouldn't make sense to add it to Animation.

animation-proposal.md
eanders-MS: I like the renaming, but unsure how `to` could be optional, since interpolations need a target value. Can you expand?

animation-proposal.md
eanders-MS: `from` and `to` are both required parameters, unless you're cancelling an existing interpolation (see examples section).

animation-proposal.md
```diff
@@ -0,0 +1,338 @@
...
+ * up the value client-side.
+ */
+public selector(path: string): string;
+```
+
+
+#### Convenience API
+
+An `interpolate` method on context isn't very discoverable. We may want to add convenience methods on `Actor` and `Asset`, so that `interpolate` shows up in intellisense. Example:
+
```
sorenhan: Why have this as a "convenience API", and not just leave this as the only option? Why would anyone want to use context.interpolate(actor.selector())?. That may be how it works internally, but there's no benefit for end users to worry about that, right?

animation-proposal.md
eanders-MS: Actually, I think nlerp may be more appropriate, and performant.

animation-proposal.md
eanders-MS: Yep that would be clearer/simpler.

animation-proposal.md
eanders-MS: I see now. Good suggestion.

animation-proposal.md
eanders-MS: Should be able to use the initial value of the field referenced by `selector` as the initial `from` value.

animation-proposal.md
eanders-MS: Added to Actor and Asset, and removed from Context.

eanders-MS: Actually, there are some issues with moving this method off Context. Reverting that design change for the moment.

animation-proposal.md
eanders-MS: Turns out there is an issue with moving `interpolate` off `Context`. Reverting that change for now and let's chat in person.

animation-proposal.md
```diff
@@ -0,0 +1,365 @@
...
+```ts
+// Play the 'jump' animation, then do something when it's done.
+actor.animation("jump").start().then(() => console.log("done jumping"));
+
+// Play the 'victory' animation, and wait for it to complete.
+await actor.animation("victory").start();
+```
+
+
+### Interpolations
```
stevenvergenz: Would be good to support an animation path in this API too. So not just lerping, but following a 3d Bezier curve. This would be needed for common simple animations like a bounce or an arc. Could be implemented by setting the `to` property to an array.

eanders-MS: Great idea.

PR #388, opened 2019-07-19
Soren/fix mediainstance timing

sorenhan:
This removes an unnecessary parameter (startTimeOffset - not needed when we have a time variable in setmediastateoptions), repairs the time calculation on join-in-progress (was broken if using seek, and if doing more than one join-in-progress).

### DIFF
--- a/packages/functional-tests/src/tests/sound-test.ts
+++ b/packages/functional-tests/src/tests/sound-test.ts
@@ -82,9 +82,9 @@ export default class SoundTest extends Test {
 				looping: true,
 				doppler: 0.0,
 				spread: 0.7,
-				rolloffStartDistance: 2.5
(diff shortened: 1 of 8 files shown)


packages/functional-tests/src/tests/sound-test.ts
```diff
@@ -82,9 +82,9 @@ export default class SoundTest extends Test {
 				looping: true,
 				doppler: 0.0,
 				spread: 0.7,
-				rolloffStartDistance: 2.5
-			},
-			0.0);
+				rolloffStartDistance: 2.5,
+				time: 30.0
```
stevenvergenz: Was the time changed deliberately? 0 seconds vs 30 seconds?

PR #406, opened 2019-07-28
user-update strings were named wrong, I believe

SorenHannibal:
I spotted that these were not named right, so I think our user-update message may not have worked properly. I haven't tested this behavior after spotting is, so we should review this part of the code

### DIFF
--- a/packages/sdk/src/types/internal/context.ts
+++ b/packages/sdk/src/types/internal/context.ts
@@ -513,7 +513,7 @@ export class InternalContext {
 		if (isNewUser) {
 			this.context.emitter.emit('user-joined', user);
 		} else {
-			this.context.emitter.emit('user-updated', user);
+			this.context.emitter.emit('user-update', user);
 		}
 	}
 
--- a/packages/sdk/src/types/internal/context.ts
+++ b/packages/sdk/src/types/internal/context.ts
@@ -345,7 +345,7 @@ export class InternalContext {
 			execution.on('protocol.destroy-actors', this.localDestroyActors);
 			execution.on('protocol.user-joined', this.userJoined);
 			execution.on('protocol.user-left', this.userLeft);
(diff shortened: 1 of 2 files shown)


stevenvergenz:
It worked before, because `user-mask-test` passes. All you've done here is change the name of the event, and change the subscription to match. It doesn't really matter what the event name is, as long as they match.

SorenHannibal:
ah, you are right. I didn't see that protocol.update-user was not related to user-update. 
That said, I found another place where we had user-updated instead of user-update

packages/sdk/src/types/internal/context.ts
```diff
@@ -513,7 +513,7 @@ export class InternalContext {
 		if (isNewUser) {
 			this.context.emitter.emit('user-joined', user);
 		} else {
-			this.context.emitter.emit('user-updated', user);
+			this.context.emitter.emit('user-update', user);
```
eanders-MS: >user-update [](start = 30, length = 11)

`user-updated` was correct. (This is different from the event you're renaming above).

packages/sdk/src/types/internal/context.ts
```diff
@@ -345,7 +345,7 @@ export class InternalContext {
 			execution.on('protocol.destroy-actors', this.localDestroyActors);
 			execution.on('protocol.user-joined', this.userJoined);
 			execution.on('protocol.user-left', this.userLeft);
-			execution.on('protocol.update-user', this.updateUser);
+			execution.on('protocol.user-update', this.updateUser);
```
eanders-MS: >user-update [](start = 26, length = 11)

Even though this rename is essentially a noop (it worked before), I like user-update better than update-user, since user-update matches the payload type of the same name.

sorenhan:
Abandoning this, it doesn't add any value

Issue #24, opened 2018-12-05
Send MREAPI.Logger.LogXXXXX() text through to Node.js 


sorenhan:
When an MRE is instantiated, and the client has errors, the errors should be sent back to the MRE, so they can be piped through to where developers expect it. By default I would expect it to show up in the Visual Studio Debug Console, and in Node.js' log files.

Examples where I would expect errors to show up
If an MRE client tries to play an animation that doesn't exist on an actor
If MRE fail to load a GLTF
If MRE tries to access an actor that doesn't exist

eanders-MS:
Add an onError handler to context.

eanders-MS:
Design: Add a "log-message" payload that can be sent any time from the client. Just need to log it (no onError handler). Log it to the "engine" facility.

sorenhan:
Changed description to highlight that we already have MRE Unity logging APIs, we just don't push them through

PR #407, opened 2019-07-28
moved client traces into a separate logging facility

SorenHannibal:
Instead of client library traces going into the "network" category, they have their own category, and are enabled by default in functional tests. This is part of #24 , and builds on the logging PR on the unity repo.

### DIFF
--- a/packages/functional-tests/src/server.ts
+++ b/packages/functional-tests/src/server.ts
@@ -14,8 +14,6 @@ dotenv.config();
 process.on('uncaughtException', (err) => console.log('uncaughtException', err));
 process.on('unhandledRejection', (reason) => console.log('unhandledRejection', reason));
 
-MRE.log.enable('app');
-MRE.log.enable('client-trace');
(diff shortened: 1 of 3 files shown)


packages/functional-tests/src/server.ts
```diff
@@ -15,6 +15,7 @@ process.on('uncaughtException', (err) => console.log('uncaughtException', err));
 process.on('unhandledRejection', (reason) => console.log('unhandledRejection', reason));
 
 MRE.log.enable('app');
+MRE.log.enable('client-trace');
```
eanders-MS: >client-trace [](start = 16, length = 12)

What about just `client`? client-trace seems overly long, and `trace` might be misleading/open to misinterpretation.

eanders-MS: >client [](start = 16, length = 6)

Let's enable this by default (done in log.ts checkInitialize method). By the way, 'app' is enabled by default now, no longer necessary to enable it here.

PR #409, opened 2019-07-31
Mesh design discussion

stevenvergenz:


mesh-design.md
```diff
@@ -0,0 +1,270 @@
...
+complicating the API. See [Other Considerations](#consolidating-the-collider-apis-into-actor-patch).
+
+For example, if I wanted to create a box button, I might do something like this:
+
+```ts
+const context: MRE.Context;
+const assets: MRE.AssetContainer;
+
+// With a mesh asset and auto collider:
+const actor = MRE.Actor.CreateEmpty(context, {
```
eanders-MS: The name `Actor.CreateEmpty` doesn't work anymore. How about just `Actor.Create`?

mesh-design.md
```diff
@@ -0,0 +1,270 @@
...
+
+3. Through an Actor instance's `setCollider(colliderType: 'box' | 'sphere', ...moreOptions)` method, which takes
+	the shape, size, and options of the desired collider, and adds it to the accumulating actor patch. Next
+	`actor-update`, that collider patch is converted into a Unity collider of the appropriate type.
+
+**Note**: Mesh colliders in Unity are [expensive to generate, expensive to evaluate, and have a laundry list of
+caveats](https://docs.unity3d.com/Manual/class-MeshCollider.html). For this reason they generally advise against
+adding mesh colliders at all at runtime. So if we want MREs to have the ability to produce mesh colliders at all,
+we should be prepared for those performance penalties. For this reason, only `CreateFromGltf` can generate mesh
+colliders, though I suspect this isn't great either.
```
eanders-MS: Let's do what we talked about at lunch and allow 'mesh' colliders to be generated everywhere and just document that it's costly. The fact that it's not performant is a Unity limitation, but may not apply to future hosts (Babylon, etc).

tombuMS: I do hope lunch was tasty and filled with an enjoyable conversation about mesh colliders.  Sorry I couldn't attend the meeting due to my location being in NJ.  What was the conversation about and how did we come to this conclusion?

mesh-design.md
```diff
@@ -0,0 +1,270 @@
...
+actor.collider.geometry = {
+	shape: 'sphere',
+	radius: 0.5
+} as MRE.SphereColliderGeometry;
+```
+
+I would also recommend removing `mesh` as a shape type, so it can only be specified implicitly
+via `auto`. Or else rename `mesh` to `unmanaged` so it's clear that it can't be used. Alternatively,
+we could go in the other direction and allow mesh colliders to be created at any time, and let
+the developer determine if it's appropriate to use.
```
tombuMS: I think removing mesh in favor of auto is a better choice here and is what I would personally like to see.

mesh-design.md
```diff
@@ -0,0 +1,270 @@
...
+// Would become this:
+MRE.Actor.CreateFromPrefab(context, {
+	// new sugar: prefabId could take a promise that will resolve to an array of assets. Create and
+	// return the new local actor immediately, but postpone the create-from-prefab network message
+	// until the prefab promise resolves.
+	prefabId: assets.loadGltf(url, 'mesh'),
+	actor: {
+		name: 'gltf'
+	}
+});
```
eanders-MS: This is less discoverable than before. What about:
```
MRE.Actor.CreateFromGltf(this.assets, url, actor: { name: 'gltf' });
```

mesh-design.md
```diff
@@ -0,0 +1,270 @@
...
+	actor: { name: 'button' }
+});
+```
+
+## Other Considerations
+
+### Consolidating the collider APIs into actor patch
+
+As previously noted, `CreatePrimitive` has one way to specify a collider, `CreateFromGltf` has another,
+and `setCollider` has yet another. This setup was done specifically to make it difficult to change actor
```
tombuMS: I assume this is leaving in place the setCollider API calls on the actor for adding a collider to an actor after creation has happened?

mesh-design.md
eanders-MS: Apologies Tom! That was insensitive of me. That part of my comment was completely unnecessary, and could have simply read:
> Let's allow 'mesh' colliders to be generated everywhere and just document that it's costly. The fact that it's not performant is a Unity limitation, but may not apply to future hosts (Babylon, etc).

But I think I didn't fully grok the alternative approach that exposes 'auto'. I didn't realize it removed 'mesh' entirely. I think that's a better, simpler solution. I retract my earlier recommendation of exposing 'mesh' everywhere in favor of 'auto'.

mesh-design.md
stevenvergenz: Correct, to match the other components that go through the patch: initialization through the Like interfaces, and modification through the dedicated API.

---
In reply to: [311278852](https://github.com/microsoft/mixed-reality-extension-sdk/pull/409#discussion_r311278852) [](ancestors = 311278852)

mesh-design.md
stevenvergenz: It turns out this might be necessary regardless of my opinion on it. Turns out I can't arbitrarily postpone the message as I described here without also postponing every subsequent message that pertains to the created actor (set-behavior, set-sound-state, etc.). It's doable in the sync layer, but this seems controversial. Let's have a meeting to discuss.

---
In reply to: [311278822](https://github.com/microsoft/mixed-reality-extension-sdk/pull/409#discussion_r311278822) [](ancestors = 311278822)

Issue #239, opened 2019-03-04
Cleanup old sound instances in sync layer


eanders-MS:
Not critical.

sorenhan:
Was accidentally marked as fixed - we only fixed this in the unity side, not the sync layer

sorenhan:
There are 3 separate cases to deal with:
1. non-looping sounds expire after a certain amount of time - should clean themselves up
2. when an actor is freed, the sound instances should be freed, too
3. when a sound asset is freed, it should find and free those sound isntances, too

PR #412, opened 2019-08-02
Soren/cleanup expired sounds

SorenHannibal:
Fixes the most important parts of #239 - garbage collect mediainstances for an actor whenever the actor tries to add a new mediainstance. And aggressively filtering out before clientsyncing.
And they are all destroyed when the actor is destroyed, too

This does not fix the case where an asset is destroyed - we ought to send a Stop command then, but it's much less important

### DIFF
--- a/packages/sdk/src/adapters/multipeer/protocols/clientSync.ts
+++ b/packages/sdk/src/adapters/multipeer/protocols/clientSync.ts
@@ -274,6 +274,11 @@ export class ClientSync extends Protocols.Protocol {
 				// TODO This sound tweaking should ideally be done on the client, because then it can consider the
 				// time it takes for packet to arrive. This is needed for optimal timing .
 				const targetTime = Date.now() / 1000.0;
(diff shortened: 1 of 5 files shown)


packages/sdk/src/types/internal/context.ts
```diff
@@ -309,7 +309,7 @@ export class InternalContext {
 			execution.on('protocol.destroy-actors', this.localDestroyActors);
 			execution.on('protocol.user-joined', this.userJoined);
 			execution.on('protocol.user-left', this.userLeft);
-			execution.on('protocol.update-user', this.updateUser);
+			execution.on('protocol.user-update', this.updateUser);
```
eanders-MS: >user-update [](start = 26, length = 11)

The user-update system isn't in use, and will need revamping if/when it does start getting used, but I'm not sure the changes you made here are an improvement over what was there already. I'd feel better these changes were reverted.

sorenhan: fixed

PR #415, opened 2019-08-13
Mesh support

stevenvergenz:
* `loadGltf` now returns Mesh assets
* `CreatePrimitive` now lives on `AssetContainer`, and returns a mesh asset
* Add `meshId` field to actor appearance objects
* Remove `Actor.CreateFromGltf`; use `CreateFromPrefab` instead
* Rename `Actor.CreateEmpty` to simply `Actor.Create`
* Add `auto` collider type
* Make collider properties patchable


packages/sdk/src/types/runtime/actor.ts
```diff
@@ -250,30 +224,15 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
 	 * Creates a new actor hierarchy from the provided prefab.
 	 * @param context The SDK context object.
-	 * @param options.prefabId The ID of the prefab asset.
+	 * @param options.prefabId The ID of the prefab asset, or an array of assets from which the first prefab is taken.
 	 * @param options.actor The initial state of the root actor.
 	 * given a collider type when loaded @see AssetManager.loadGltf.
 	 */
 	public static CreateFromPrefab(context: Context, options: {
-		prefabId: string,
+		prefabId: string | Asset[],
```
eanders-MS: `prefabId` with type Asset[] is odd. How about creating an overload for it?

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -346,7 +305,7 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 	 * @param colliderType Type of the collider to enable.
 	 * @param isTrigger Whether the collider is a trigger volume or not.
 	 * @param center The center of the collider, or default of the object if none is provided.
-	 * @param radius The radius of the collider, or default bounding if non is provided.
+	 * @param radius The radius of the collider.
```
eanders-MS: What is the default radius? Would be good to mention it in the doc comment.

packages/functional-tests/src/tests/grab-test.ts
```diff
@@ -36,12 +43,10 @@ export default class GrabTest extends Test {
...
 		});
 
-		// Load a glTF model
-		this.model = MRE.Actor.CreateFromGltf(this.app.context, {
-			// at the given URL
-			resourceUrl: `${this.baseUrl}/monkey.glb`,
-			// and spawn box colliders around the meshes.
-			colliderType: 'box',
+		// Create an actor
+		this.model = MRE.Actor.CreateFromPrefab(this.app.context, {
```
eanders-MS: Simply loading a gltf is 2x more expensive now, requiring an additional round trip message. It's also more We need to find a way to move the async part of this operation to the client.

stevenvergenz: If you're only looking at network traffic that's true. I guess I could add an `AssetContainer.loadAndSpawnGltf` method?

---
In reply to: [313511937](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313511937) [](ancestors = 313511937)

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -361,7 +320,7 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 	 * @param colliderType Type of the collider to enable.
 	 * @param isTrigger Whether the collider is a trigger volume or not.
 	 * @param center The center of the collider, or default of the object if none is provided.
-	 * @param size
+	 * @param size The dimensions of the collider.
```
tombuMS: >The dimensions of the collider. [](start = 16, length = 31)

Should add in default size here inline with the above comment.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -370,22 +329,45 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+	 */
+	public setCollider(
+		colliderType: 'auto',
+		isTrigger: boolean
+	): void;
+
 	/** @ignore */
 	public setCollider(
-		colliderType: 'box' | 'sphere',
+		colliderType: 'box' | 'sphere' | 'capsule' | 'auto',
```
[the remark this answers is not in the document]
tombuMS: Yes.  This should be the actual type that we have for this.

---
In reply to: [313499840](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313499840) [](ancestors = 313499840)

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -733,24 +715,33 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
 					center,
 					radius: size as number
 				} as SphereColliderGeometry;
-
+			case 'capsule':
+				return {
+					shape: 'capsule',
+					center,
+					size: size as Partial<Vector3Like>
+				} as CapsuleColliderGeometry;
```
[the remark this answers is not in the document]
tombuMS: When I originally wrote this, it gave errors without the explicit typing here.  If that is not the case anymore for whatever reason, we could remove it for simplification.

---
In reply to: [313500764](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313500764) [](ancestors = 313500764)

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -761,18 +752,18 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 	}
 
 	private _setCollider(collider: Partial<ColliderLike>) {
-		if (this._collider) {
-			unobserve(this._collider);
-			this._collider = undefined;
```
tombuMS: Why is this being removed?  We need to unobserve from the previous collider if there is one.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -761,18 +752,18 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
-		this._collider = new Collider(this, collider);
-		observe({
-			target: this._collider,
-			targetName: 'collider',
-			notifyChanged: (...path: string[]) => this.actorChanged(...path),
-			// Trigger notifications for every observed leaf node to ensure we get all values in the initial patch.
-			triggerNotificationsNow: true
-		});
+		// Copying the new values will trigger an actor update and enable/update the collider component.
+		this._collider.copy(collider);
```
tombuMS: This is actually changing the behavior of what was previously going on here and I actual believe this is incorrect.  We destroyed the old collider so that no leftover values that are not in the new collider partial are reset to default values.  Simply doing a copy here does not replicate that behavior and will produce different colliders if a value is not explicitly set from the incoming collider like, but was set in the previous collider.

packages/sdk/src/types/runtime/actor.ts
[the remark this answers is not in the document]
stevenvergenz: Good catch. Previously that type also contained 'mesh', but that's no longer the case. I'll swap that out.

---
In reply to: [313519815](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313519815) [](ancestors = 313519815,313499840)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: Enabled and IsTrigger are persisted Unity-side now, so destroying and recreating the UnityCollider keeps the values of those properties. Thus we don't need to patch them again.

---
In reply to: [313523421](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313523421) [](ancestors = 313523421)

tombuMS: We never wanted to persist those nor never did.  Why are we changing this behavior now?

---
In reply to: [313538914](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313538914) [](ancestors = 313538914,313523421)

stevenvergenz: Summary of discussions: make sure that if setCollider is called, all the fields are either specified or set to default, and make sure that we're creating the minimum number of patches.

---
In reply to: [313559093](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313559093) [](ancestors = 313559093,313538914,313523421)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: Changed the API so the radius is no longer optional. There was no reasonable default radius that I could come up with.

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: Changed the API so the size is not optional.

packages/functional-tests/src/tests/grab-test.ts
eanders-MS: Updating this thread: POR is to bring back CreateFromGltf, to preserve this convenience API.

---
In reply to: [313515597](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313515597) [](ancestors = 313515597,313511937)

packages/sdk/src/types/runtime/actor.ts
eanders-MS: How would `prefabId` work when passed an Asset array that wasn't a loaded prefab?

---
In reply to: [313508265](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313508265) [](ancestors = 313508265)

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -345,47 +304,73 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
 	 * @param center The center of the collider, or default of the object if none is provided.
-	 * @param size
 	 */
 	public setCollider(
 		colliderType: 'box',
 		// collisionLayer: CollisionLayer,
 		isTrigger: boolean,
-		center?: Vector3Like,
-		size?: Vector3Like): void;
+		size: Vector3Like,
```
tombuMS: >size: Vector3Like, [](start = 2, length = 18)

We discussed this that the Unity client provides a reasonable default for this when specified, and that optional should be allowed for the size related properties here.  These should still be optional.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -345,47 +304,73 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+	 * @param colliderType Type of the collider to enable.
+	 * @param isTrigger Whether the collider is a trigger volume or not.
+	 * @param size The dimensions of the collider, with the largest component of the vector
+	 * being the primary axis and height of the capsule, and the smallest the radius.
+	 * @param center The center of the collider, or default of the object if none is provided.
+	 */
+	public setCollider(
+		colliderType: 'capsule',
+		isTrigger: boolean,
+		size: Vector3Like,
```
tombuMS: >size: Vector3Like, [](start = 2, length = 18)

We discussed this that the Unity client provides a reasonable default for this when specified, and that optional should be allowed for the size related properties here.  These should still be optional.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -345,47 +304,73 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
-	 * @param radius The radius of the collider, or default bounding if non is provided.
 	 */
 	// * @param collisionLayer The layer that the collider operates in.
 	public setCollider(
 		colliderType: 'sphere',
 		// collisionLayer: CollisionLayer,
 		isTrigger: boolean,
-		center?: Vector3Like,
-		radius?: number): void;
+		radius: number,
```
tombuMS: >radius: number, [](start = 2, length = 15)

We discussed this that the Unity client provides a reasonable default for this when specified, and that optional should be allowed for the size related properties here.  These should still be optional.

packages/sdk/src/types/runtime/actor.ts
tombuMS: We were supposed to be allowing Unity to provide the default for sizing did we not?

---
In reply to: [314923618](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r314923618) [](ancestors = 314923618)

packages/sdk/src/types/runtime/actor.ts
tombuMS: That was not what was decided on here.  There should be an explicit reset of values in the existing collider that is no longer being replaced to set those to default.  This function accepts a Partial<ColliderLike> and should support that.  We cannot expect a fully filled out collider like to be passed in here, and thus must account for any missing properties in that collider like in this function.  What we had discussed here was:

1. Reset the existing collider to its default values or create a new collider with its default values when there is not a previous collider to reuse
2. Copy in the partial collider like.

---
In reply to: [313648641](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r313648641) [](ancestors = 313648641,313559093,313538914,313523421)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: I thought the result of that discussion was that we wanted to be explicit in the API for the sake of cross-engine consistency. I'll change it back.

---
In reply to: [315736602](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315736602) [](ancestors = 315736602)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: You said that your concern was specifically with the use of the `setCollider` APIs. As it's written, all parameters are set or are provided defaults. Isn't that the same thing?

---
In reply to: [315740022](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315740022) [](ancestors = 315740022,313648641,313559093,313538914,313523421)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: Even if that default is a constant? The more we let Unity specify defaults, the less useful intellisense becomes. Letting Unity set box collider defaults makes sense because it fits it to the mesh, but I don't think Unity computes a bounding sphere for meshes.

---
In reply to: [315736855](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315736855) [](ancestors = 315736855)

packages/sdk/src/types/runtime/actor.ts
eanders-MS: prefabId of type Asset[] is unintuitive. Create a function overload for this function signature.

---
In reply to: [315425105](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315425105) [](ancestors = 315425105,313508265)

packages/sdk/src/types/runtime/actor.ts
eanders-MS: My understand is we agreed that creating a new Collider was the right approach here, so that previous field values would not be erroneously preserved.

---
In reply to: [315898690](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315898690) [](ancestors = 315898690,315740022,313648641,313559093,313538914,313523421)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: Changed this back, and documented the defaults.

---
In reply to: [315737144](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315737144) [](ancestors = 315737144,314923618)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: Changed it back and documented defaults.

---
In reply to: [315898507](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315898507) [](ancestors = 315898507,315736602)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: Changed back, and documented default.

---
In reply to: [315736742](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315736742) [](ancestors = 315736742)

packages/sdk/src/types/runtime/actor.ts
eanders-MS: With your latest changes, the incoming collider is a Partial again.

---
In reply to: [315939018](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315939018) [](ancestors = 315939018,315898690,315740022,313648641,313559093,313538914,313523421)

packages/sdk/src/types/runtime/actor.ts
eanders-MS: This isn't what I expected. Is it not possible to create an overload? prefabId and firstPrefabFrom are mutually exclusive, correct?

---
In reply to: [315938355](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r315938355) [](ancestors = 315938355,315425105,313508265)

stevenvergenz: So you want me to create an overload that takes a slightly different options object? Overloads only deal with actual arguments, not sure what it would mean to overload an options object. It'd just be the union, which is what I have.

---
In reply to: [316358367](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r316358367) [](ancestors = 316358367,315938355,315425105,313508265)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: It always took a partial. Plus we need it to be so creating colliders in initializers isn't too verbose.

---
In reply to: [316357410](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r316357410) [](ancestors = 316357410,315939018,315898690,315740022,313648641,313559093,313538914,313523421)

packages/sdk/src/types/runtime/actor.ts
stevenvergenz: I've added some overloads, each takes one of the args. Is that right?

---
In reply to: [316405948](https://github.com/microsoft/mixed-reality-extension-sdk/pull/415#discussion_r316405948) [](ancestors = 316405948,316358367,315938355,315425105,313508265)

PR #421, opened 2019-08-20
Apply read-only asset properties after load

### LINKED ISSUE #418: Checking duration on Sound types always returns 0
It appears checking duration on a Sound always returns 0.

`        this.tracks.push(await this.assets.createSound(
            'Track1',
            { uri: `${this.baseUrl}/RadioTuningshort.wav` }
        ));

        console.log( "Duration " + this.tracks[0].duration );`

this.tracks - an array of type Sound
this.assets - an Asset Container
RadioTuningshort.wav is a 4 second clip of radio static.

Per Steven V. I have tried to make sure the Sound was created.

stevenvergenz:
Fixes #418

### DIFF
--- a/packages/sdk/src/types/runtime/assets/assetContainer.ts
+++ b/packages/sdk/src/types/runtime/assets/assetContainer.ts
@@ -188,6 +188,8 @@ export class AssetContainer {
 		if (reply.failureMessage || reply.assets.length !== 1) {
 			throw new Error(`Creation/Loading of asset ${asset.name} failed: ${reply.failureMessage}`);
 		}
+
+		asset.copy(reply.assets[0]);
 	}
 
 	private sendPayloadAndGetReply<T extends Payloads.Payload, U extends Payloads.Payload>(payload: T): Promise<U> {


packages/sdk/src/types/runtime/assets/assetContainer.ts
```diff
@@ -188,6 +188,8 @@ export class AssetContainer {
 		if (reply.failureMessage || reply.assets.length !== 1) {
 			throw new Error(`Creation/Loading of asset ${asset.name} failed: ${reply.failureMessage}`);
```
eanders-MS: Unrelated feedback, but this shouldn't be throwing an exception. Log an error instead.

stevenvergenz: This error bubbles up to rejecting the `Asset.created` promise. Isn't this what it's for?

---
In reply to: [315937817](https://github.com/microsoft/mixed-reality-extension-sdk/pull/421#discussion_r315937817) [](ancestors = 315937817)

PR #423, opened 2019-08-22
Re-add CreateFromGltf API

stevenvergenz:
Works by chaining `load-assets` and `create-from-prefab` calls under the hood. Requires a special session message so synchronous actor patches have somewhere to go while we wait for assets to load.

### DIFF
--- a/packages/sdk/src/adapters/multipeer/rules.ts
+++ b/packages/sdk/src/adapters/multipeer/rules.ts
@@ -1114,5 +1114,17 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
 				return null;
 			}
 		}
+	},
+
+	// ========================================================================
+	'x-reserve-actor': {
+		...DefaultRule,
+		session: {
+			...DefaultRule.session,
+			beforeReceiveFromApp: (session: Session, message: Message<Payloads.XReserveActor>) => {
+				session.cacheInitializeActorMessage(message);
+				return null;
+			}
+		}
 	}
 };
--- a/packages/sdk/src/adapters/multipeer/rules.ts
+++ b/packages/sdk/src/adapters/multipeer/rules.ts
@@ -1119,6 +1119,12 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
 	// ========================================================================
 	'x-reserve-actor': {
 		...DefaultRule,
+		synchronization: {
+			stage: 'never',
+			before: 'ignore',
+			during: 'ignore',
+			after: 'ignore'
+		},
 		session: {
 			...DefaultRule.session,
 			beforeReceiveFromApp: (session: Session, message: Message<Payloads.XReserveActor>) => {
--- a/packages/sdk/src/adapters/multipeer/session.ts
+++ b/packages/sdk/src/adapters/multipeer/session.ts
@@ -241,7 +241,8 @@ export class Session extends EventEmitter {
 	}
 
 	public cacheInitializeActorMessage(message: InitializeActorMessage) {
-		if (!this.actorSet[message.payload.actor.id]) {
+		const syncActor = this.actorSet[message.payload.actor.id];
+		if (!syncActor) {
 			const parent = this.actorSet[message.payload.actor.parentId];
 			this.actorSet[message.payload.actor.id] = {
 				actorId: message.payload.actor.id,
@@ -249,6 +250,16 @@ export class Session extends EventEmitter {
 					|| message.payload.actor.exclusiveToUser,
 				initialization: deepmerge({ message }, {})
 			};
+		}
+		// update reserved actor init message with something the client can use
+		else if (syncActor.initialization.message.payload.type === 'x-reserve-actor') {
+			// send real init message, but with session's initial actor state
+			message.payload = {
+				...message.payload,
+				actor: syncActor.initialization.message.payload.actor
+			}
+			// write the merged message back to the session
+			syncActor.initialization.message = message;
 		}
 	}
 
--- a/packages/sdk/src/types/internal/context.ts
+++ b/packages/sdk/src/types/internal/context.ts
@@ -39,6 +39,8 @@ import {
 	CreateFromPrefab,
 	DestroyActors,
 	InterpolateActor,
+	LoadAndSpawnPrefab,
+	LoadAndSpawnResult,
 	ObjectSpawned,
 	OperationResult,
 	Payload,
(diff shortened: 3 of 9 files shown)


packages/sdk/src/types/internal/context.ts
```diff
@@ -182,6 +164,42 @@ export class InternalContext {
...
+		container.loadGltf(options.uri, options.colliderType)
+		.then(assets => {
+			const prefab = assets.find(a => !!a.prefab);
+			if (!prefab) {
+				actor.internal.notifyCreated(false, `glTF file specifies no scenes: ${options.uri}`);
+				return;
+			}
+
+			// once assets are done, find first prefab and spawn it
+			this.CreateFromPrefab({
```
eanders-MS: >CreateFromPrefab [](start = 8, length = 16)

Problem: CreateFromPrefab is going to create a new Actor with a different id.

packages/sdk/src/adapters/multipeer/session.ts
```diff
@@ -241,14 +241,24 @@ export class Session extends EventEmitter {
...
 			this.actorSet[message.payload.actor.id] = {
 				actorId: message.payload.actor.id,
 				exclusiveToUser: parent && parent.exclusiveToUser
 					|| message.payload.actor.exclusiveToUser,
 				initialization: deepmerge({ message }, {})
 			};
+		// update reserved actor init message with something the client can use
+		} else if (syncActor.initialization.message.payload.type === 'x-reserve-actor') {
+			// send real init message, but with session's initial actor state
+			message.payload = {
```
eanders-MS: >message [](start = 3, length = 7)

Safer to make a deepmerge copy of message here before modifying it.

packages/sdk/src/types/internal/context.ts
stevenvergenz: Damn, you're right. I think everything works as is, except for the UUID. I'll modify it to only add a new ID if it's missing one.

---
In reply to: [316824146](https://github.com/microsoft/mixed-reality-extension-sdk/pull/423#discussion_r316824146) [](ancestors = 316824146)

packages/sdk/src/adapters/multipeer/session.ts
stevenvergenz: I'm not sure it does in context. The actor patch in the create message is a strictly older version of the one in the session, I don't think we want anything from it.

---
In reply to: [316825178](https://github.com/microsoft/mixed-reality-extension-sdk/pull/423#discussion_r316825178) [](ancestors = 316825178)

packages/sdk/src/types/internal/context.ts
```diff
@@ -103,13 +103,15 @@ export class InternalContext {
...
-				...options.actor,
-				id: UUID()
-			}
+				...options.actor
+			},
+			type: 'create-from-prefab'
 		} as Payloads.CreateFromPrefab;
+		if (!payload.actor.id) {
+			payload.actor.id = UUID();
+		}
```
eanders-MS: I don't think this is safe. What if I want to use an existing actor as a template for a new one?

packages/sdk/src/types/internal/context.ts
eanders-MS: I don't think this is fixed yet. Maybe make a private method they both call?

---
In reply to: [316851109](https://github.com/microsoft/mixed-reality-extension-sdk/pull/423#discussion_r316851109) [](ancestors = 316851109,316824146)

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -282,9 +293,12 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
-	 * @param container
-	 * @param options
+	 * @deprecated
+	 * Use [[AssetContainer.loadGltf]] and [[Actor.CreatePrefab]] instead.
+	 * @param container The asset container to load the glTF assets into
+	 * @param options.uri A URI to a .gltf or .glb file
+	 * @param options.colliderType The type of collider to add to each mesh
+	 * @param options.actor The initial state of the actor
 	 */
 	public static CreateFromGltf(container: AssetContainer, options: {
```
eanders-MS: Great idea to deprecate methods rather than delete them, but don't deprecate `CreateFromGltf`. It's a convenience method we should keep.

packages/sdk/src/types/runtime/actor.ts
```diff
@@ -294,6 +308,35 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
 
+	/**
+	 * @deprecated
+	 * Use AssetContainer.create*Mesh and [[Actor.Create]] instead.
+	 * @param container The asset container to load the mesh into
+	 * @param options.definition The primitive shape and size
+	 * @param options.addCollider Add an auto-typed collider to the actor
+	 * @param options.actor The initial state of the actor
+	 */
+	public static CreatePrimitive(container: AssetContainer, options: {
```
eanders-MS: Let's not deprecate this method either. It's very convenient and discoverable for non-power-users.

packages/sdk/src/types/internal/context.ts
stevenvergenz: That private method they both call *is* createActorFromPayload. All the individual Create calls do is build the payload. I looked into it, but there is no shared code between CreateFromPrefab and this if it's not generating a UUID.

---
In reply to: [316930810](https://github.com/microsoft/mixed-reality-extension-sdk/pull/423#discussion_r316930810) [](ancestors = 316930810,316851109,316824146)

packages/sdk/src/adapters/multipeer/rules.ts
```diff
@@ -1114,5 +1114,23 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
 				return null;
 			}
 		}
+	},
+
+	// ========================================================================
+	'x-reserve-actor': {
```
tombuMS: >'x-reserve-actor' [](start = 1, length = 17)

I get what you are doing here in trying to name this and identify it as a exception type, but maybe we could be more descriptive here.  How about 'session.reserve-actor'?

packages/sdk/src/types/internal/context.ts
eanders-MS: However it's accomplished, we need to make certain that application code cannot specify or override the new actor's id.

---
In reply to: [317196273](https://github.com/microsoft/mixed-reality-extension-sdk/pull/423#discussion_r317196273) [](ancestors = 317196273,316930810,316851109,316824146)

packages/sdk/src/adapters/multipeer/rules.ts
eanders-MS: My take: I like this name. It's not overly specific, it's inline with custom http header naming conventions, and alphabetically sorts toward the end of a list.

---
In reply to: [317267483](https://github.com/microsoft/mixed-reality-extension-sdk/pull/423#discussion_r317267483) [](ancestors = 317267483)

packages/sdk/src/adapters/multipeer/session.ts
eanders-MS: I vaguely recall that making a deep copy of the entire message was important to do here, but don't remember the reason now. We should preserve that behavior.

packages/functional-tests/src/tests/gltf-gen-test.ts
```diff
@@ -41,10 +41,8 @@ export default class GltfGenTest extends Test {
 			})]
 		})]);
 
-		MRE.Actor.CreateFromPrefab(this.app.context, {
-			firstPrefabFrom: await this.assets.loadGltf(
-				Server.registerStaticBuffer('sphere.glb', gltfFactory.generateGLTF())
-			),
+		MRE.Actor.CreateFromGltf(this.assets, {
+			uri: Server.registerStaticBuffer('sphere.glb', gltfFactory.generateGLTF()),
```
eanders-MS: >GLTF [](start = 70, length = 4)

Unrelated, but inconsistent naming here: Rename this `generateGltf`?

PR #425, opened 2019-08-22
Initial SDK-side work for client-triggered actions

eanders-MS:
Had to do some weird context plumbing, but it's all internal. Lemme know if you can think of something better.


packages/sdk/src/types/runtime/behaviors/clientAction.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export class PlayAnimationAction extends ClientAction {
+	constructor(private animationName: string) {
+		super();
+	}
+
+	/** @hidden */
+	public toJSON() {
+		return {
+			type: 'play-animation',
```
tombuMS: The type needs to be outside of the json here.  We need to know the type in the client before deserializing this blob, because we have to allocate the appropriate class to fill in.

packages/sdk/src/types/runtime/behaviors/clientAction.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export abstract class ClientAction {
+	/** @hidden */
+	public abstract toJSON(): any;
+
+	/** @hidden */
+	public sendPayload(context: Context, targetId: string, actionName: string, actionState: string) {
+		context.internal.sendPayload({
+			type: 'set-client-action',
+			clientAction: {
```
tombuMS: This is the level where we would need to know the client action type to allocate the C# object for Newtonsoft to populate.

packages/sdk/src/types/runtime/behaviors/action.ts
```diff
@@ -24,13 +29,25 @@ export class DiscreteAction {
 	private handlers: ActionHandlers = {};
 	private activeUserIds: string[] = [];
 
+	/** @hidden */
+	constructor(private name: string) {
```
tombuMS: I think maybe the DiscreteAction class should get a reference back to the behavior that owns it.  That would get rid of the awkward plumbing in the `on` call below that doesn't feel right.  I had already felt like a reference back would have been good here, but I didn't initially need it before.  Seems like a good time to add it here.

packages/sdk/src/types/runtime/behaviors/action.ts
```diff
@@ -24,13 +29,25 @@ export class DiscreteAction {
...
+
 	/**
 	 * Add a handler for the given action state for when it is triggered.
 	 * @param actionState The action state that the handle should be assigned to.
 	 * @param handler The handler to call when the action state is triggered.
 	 */
-	public on(actionState: ActionState, handler: ActionHandler): this {
+	public on(
+		context: Context,
+		actorId: string,
```
tombuMS: These arguments seem very awkward and out of place here.  Can't we simply pass in a reference to the behavior that owns this action, and thus has access to the context and actor ID from there?  Seems cleaner.

packages/sdk/src/types/runtime/behaviors/clientAction.ts
eanders-MS: Untrue. In the deserializer, look at JObject['type'] before reading the rest of the values out of the JObject. This is how the CollisionGeometryConverter works.

---
In reply to: [316824678](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316824678) [](ancestors = 316824678)

packages/sdk/src/types/runtime/behaviors/clientAction.ts
eanders-MS: I think you will have what you need in the deserializer. There will always be a JObject['clientAction'], and it will always have a JObject['type'].

---
In reply to: [316824978](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316824978) [](ancestors = 316824978)

packages/sdk/src/types/runtime/behaviors/action.ts
eanders-MS: Are DiscreteActions specific to behaviors? They seem generic enough they could work for colliders and other kinds of things that generate events.

---
In reply to: [316826396](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316826396) [](ancestors = 316826396)

packages/sdk/src/types/runtime/behaviors/clientAction.ts
tombuMS: Yeah.  I misread the converter code.  Wrote that a while ago.

packages/sdk/src/types/runtime/behaviors/clientAction.ts
tombuMS: As I mentioned above.  Misread the converter code that I wrote a while back.  Disregard.

packages/sdk/src/types/runtime/behaviors/action.ts
tombuMS: Discrete actions are designed specifically for the behavior system.  Handling of trigger and collision events are a bit different and contain different data.  Adding in behavior reference here is the right thing to do I believe.

tombuMS: If we decide to re-purpose this, we would likely have base class and subclass relationships for them anyways.

packages/sdk/src/types/runtime/behaviors/clientAction.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { Context } from '..';
+import { SetAudioStateOptions } from '../../..';
+import * as Payloads from '../../network/payloads';
+
+// tslint:disable: max-classes-per-file
+
+export abstract class ClientAction {
```
tombuMS: I would rename this `ClientActionHandler`.  Actions and handlers are very different concepts in the behavior system, and these are handlers for actions.  Not new types of actions

packages/sdk/src/types/network/payloads/payloads.ts
```diff
@@ -321,6 +322,11 @@ export type SetBehavior = Payload & {
 	behaviorType: BehaviorType;
 };
 
+export type SetClientAction = Payload & {
+	type: 'set-client-action';
+	clientAction: any;
```
tombuMS: Should this be a union type of the client action handlers we are adding so that this is a bit more strongly typed?  That is what we do with the collider geometry types.

packages/sdk/src/types/runtime/behaviors/action.ts
eanders-MS: How would grab work in this case? The grab DiscreteAction lives directly on Actor.

---
In reply to: [316832001](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316832001) [](ancestors = 316832001)

eanders-MS: That makes me wonder if grab should be moved to a GrabBehavior.

---
In reply to: [316840866](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316840866) [](ancestors = 316840866,316832001)

tombuMS: Ah yes.  I forgot about grab.  Yeah.  I guess the `on` function will need to take those parameters.  Just seems awkward.  Grab was not a behavior intentionally as from the perspective of the SDK, we want grabbability to be a characteristic of actors all up.  Not whether they have a behavior on them or not.

packages/sdk/src/types/runtime/behaviors/clientAction.ts
eanders-MS: I don't love this name, but "action" seems more accurate than "handler" to my mind. They don't handle anything. They just do something. They're actions. We've already defined "behavior", "action", and "handler". Do we need a fourth term for what these are?

---
In reply to: [316833692](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316833692) [](ancestors = 316833692)

packages/sdk/src/types/network/payloads/payloads.ts
eanders-MS: Yes I was being lazy. It's a little different here because clientAction is a `class` rather than a `type`, so `ClientAction` is the base class rather than a union type. But it should work.

---
In reply to: [316834322](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316834322) [](ancestors = 316834322)

eanders-MS: Or, I could make it a type instead. That might be more convenient, and inline with what we do elsewhere.

---
In reply to: [316847203](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316847203) [](ancestors = 316847203,316834322)

packages/sdk/src/types/runtime/behaviors/clientAction.ts
tombuMS: They are handling the action that is firing on the behavior.  To call them an action is not inline with what we currently refer to as an action in the SDK.

tombuMS: They are more of a reaction to something.  Maybe refer to them as a `ClientReaction`?

eanders-MS: How about `TriggeredAction`?

---
In reply to: [316852291](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316852291) [](ancestors = 316852291)

packages/sdk/src/types/network/payloads/payloads.ts
tombuMS: Whichever you prefer.  I just like having stuff strongly typed like this.

packages/sdk/src/types/runtime/behaviors/clientAction.ts
tombuMS: That is fine by me.  I just don't like it being called a ClientAction, as the behavior actions are client actions.  I think vocabulary is important here, and is why I am being a bit picky.

tombuMS: I actually really like that name now that I think about it.  Let's go with that.  Very descriptive.

packages/sdk/src/types/network/payloads/payloads.ts
eanders-MS: Made it a type union.

---
In reply to: [316855457](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316855457) [](ancestors = 316855457)

packages/sdk/src/types/runtime/behaviors/action.ts
```diff
@@ -24,13 +30,31 @@ export class DiscreteAction {
...
+		options: ActionHandler | ActionHandlerWithTriggeredAction): this {
+		const handler = (typeof options === 'function') ? options : options.handler;
 		this.handlers[actionState] = handler;
+		if (typeof options === 'object') {
+			context.internal.sendPayload({
+				type: 'set-triggered-action',
+				targetId: actorId,
+				actionName: this.name,
+				actionState,
+				action: options.action,
```
eanders-MS: >action [](start = 4, length = 6)

I suppose this should be called `triggeredAction` to differentiate meaning from `actionName` and `actionState`, which use the term "action" differently.

packages/sdk/src/types/runtime/behaviors/action.ts
tombuMS: I guess we have to live with this here.  Not the worst thing in the world.  Maybe a better way to refactor this in the future.

packages/sdk/src/types/network/payloads/payloads.ts
tombuMS: Thanks.

packages/sdk/src/types/runtime/behaviors/action.ts
eanders-MS: Glad you're ok with it, at least temporarily. It's all internal thankfully. I think we should mark this method @hidden.

---
In reply to: [316876865](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316876865) [](ancestors = 316876865)

packages/sdk/src/types/runtime/behaviors/action.ts
```diff
@@ -52,7 +52,7 @@ export class DiscreteAction {
 				targetId: actorId,
 				actionName: this.name,
 				actionState,
-				action: options.action,
+				triggeredAction: options.action,
```
tombuMS: Should the naming here be options.triggeredAction here as well for consistency?

packages/sdk/src/types/runtime/behaviors/action.ts
tombuMS: I agree.

packages/sdk/src/types/runtime/behaviors/action.ts
eanders-MS: Yes, good catch.

---
In reply to: [316894771](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316894771) [](ancestors = 316894771)

eanders-MS: Renamed.

---
In reply to: [316896105](https://github.com/microsoft/mixed-reality-extension-sdk/pull/425#discussion_r316896105) [](ancestors = 316896105,316894771)

PR #426, opened 2019-08-22
[SDK] Basic support for "triggered actions"

eanders-MS:
Triggered actions are actions that run immediately in response to an input event. e.g.: "When the user clicks this actor, run this animation".

This is a partial implementation, just enough to experiment with. Will expand to other event types (collision, notably) and additional action types once we feel we have a good base.

Will submit an RFC/spec retroactively. Will not merge this to red until it goes through spec review.


eanders-MS:
TODO: Add to functional test. I'm thinking to add it to the functional test menu itself.

eanders-MS:
TODO: Add a sample for this feature.

packages/functional-tests/src/tests/triggered-action-test.ts
```diff
@@ -0,0 +1,53 @@
...
+				triggeredAction: {
+					type: 'play-animation',
+					animationName: 'animation:0'
+				}
+			})
+			.onHover('exit', {
+				triggeredAction: {
+					type: 'stop-animation',
+					animationName: 'animation:0'
+				}
```
tombuMS: We should test one here that takes the `actorId` property that would fire the sound/animation on a different actor than the one that the triggered action is attached to.

packages/sdk/src/types/runtime/behaviors/action.ts
```diff
@@ -24,13 +25,34 @@ export class DiscreteAction {
...
+	 * @hidden
 	 * Add a handler for the given action state for when it is triggered.
+	 * @param actorId The actor associated with this action.
 	 * @param actionState The action state that the handle should be assigned to.
 	 * @param handler The handler to call when the action state is triggered.
 	 */
-	public on(actionState: ActionState, handler: ActionHandler): this {
+	public on(
+		sendPayload: (payload: Payloads.Payload) => void,
+		actorId: string,
```
tombuMS: I like this solution better than the whole passing in the context.  This does seem to be becoming more of a pattern where we want to send specific payloads from elsewhere besides in the actor and the context itself.  Seems we might want to consider a more elegant solution in the long run.  Thoughts?

packages/sdk/src/types/runtime/behaviors/triggeredAction.ts
```diff
@@ -0,0 +1,75 @@
...
+	/**
+	 * The name of the animation to stop.
+	 */
+	animationName: string,
+	/**
+	 * The actor on which to stop the animation. If omitted, the animation plays on the actor associated with
+	 * the event that fired.
+	 */
+	targetId?: string
+};
```
sorenhan: Love triggeredactions for sounds and effects, and anything that generally can't mess up synchronization. 
I think we could allow animations on single-user actors as well, but I am afraid that allowing triggeraction animations on regular actors can cause too desyncs if developers don't watch out. And in general we try to protect developers from "be careful with this" kind of feature, so I'd rather restrict it to known safe places.

I'd think that once we refactor animation to run similar to mediainstances, we would be able to mark a subset of animations (the ones that animate parameters such as color etc.) to be safe, whereas we can block changing anything that affects the physics simulation (transform, enable/disable, collision, rigid body parameters).

Issue #135, opened 2019-01-25
Modal text/String Input


sorenhan:
The functionality would probably come in as a component of any grand unifying UI system, but maybe we could deliver a standalone solution sooner? 

Maybe a host-app independent interface for "enter a text string", where the host app can decide to show a virtual keyboard (with text entry field), or pop up a messagebox itself.

willneedit:
`User.showKeyboard(defaultInput?: string): Promise<string>;`

comes to my mind. This function could pop up the virtual keyboard, optionally pre-filled with some content, and the resolved promise would hold the actual user input, or the unchanged default.

Rejected promises could come with an error code denoting the reason, where 'cancelled' might be the most obvious one, but still leaves room for cases where it's not feasible to show a keyboard (because it is already shown because of another text input, a dialog is open, ...)

eanders-MS:
Hi @willneedit I was thinking along similar lines, though wanted to include a prompt title as well:
```ts
User.textPrompt(options?: {
    title?: string, // dialog prompt
    value?: string // prefilled value
}): Promise<string>;
```
I'd like to not have to wrap every call to textPrompt in a try/catch to handle user-cancellation which really is a non-error. Probably better to return null if user cancels:
```ts
const username = await user.textPrompt({ title: "What's your name?" });
if (username) {
    console.log(`Hello ${username}`);
} else {
    console.log("user cancelled");
}
```

willneedit:
null, as opposed to an empty string, which someone could voluntarily enter. Good catch, and makes things lot easier.

PR #444, opened 2019-09-03
[Design] Dialog system

stevenvergenz:
Will address #135


dialog_design.md
```diff
@@ -0,0 +1,68 @@
...
+	to a subset of users with this option. Takes either a `GroupMask`, a `User[]`, or just one `User`. This filtering
+	could be implemented in the sync layer fairly easily, I think.
+4. `icon` (optional) - A graphic to display alongside the text. Can be one of the built-in icons (`info`, `warning`,
+	`error`, or `question`), or a custom texture supplied by the developer.
+5. `persistent` (optional) - If `true`, the dialog is presented to users in the recipients list even if they are
+	not currently online. I imagine this would be implemented by an extra handler to `context.onUserJoined`, with
+	special logic to not send the dialog to the same user more than once. This is complicated by changing membership
+	in groups, not sure how to handle that yet.
+
+Once the dialog is set up, you would call `dialog.send()` to deliver the messages to clients. If you need to hide
```
eanders-MS: Let's rename this `show()`. Up until now we haven't emphasized that anything about this API is network-centric.

packages/sdk/src/types/runtime/dialog.ts
```diff
@@ -0,0 +1,86 @@
...
+	/** A yellow exclamation point */
+	Warning,
+	/** A red exclamation point */
+	Error,
+	/** A question mark */
+	Question
+}
+
+/** Pop a message to users, and get a response */
+export class Dialog extends EventEmitter {
```
eanders-MS: I think a fluent API design makes sense here:
```ts
const dialog = new Dialog(this.context)
    .type('info')
    .text("You lost the game.")
    .recipients(loser)
    .on('response', () => {
       console.log('User acknowledged.');
    });

dialog.show();
```

And a convenience method to avoid having to call the constructor:

```ts
Dialog.Create(this.context)
    .type('info')
    .text("You lost the game.")
    .recipients(loser)
    .on('response', () => {
       console.log('User acknowledged.');
    })
    .show();
```

packages/sdk/src/types/runtime/dialog.ts
```diff
@@ -0,0 +1,86 @@
...
+	/** A dialog with text, an icon, and an OK button */
+	Message = 0,
+	/** A dialog with text, an icon, and OK and Cancel buttons */
+	Confirm,
+	/** A dialog with prompt text, a text input field, and OK and Cancel buttons */
+	Prompt
+}
+
+
+export enum DialogIcon {
```
eanders-MS: Make this a `type` and use string values:
```ts
export type DialogIcon
    = 'info'
    | 'warning'
    | 'error'
    | 'question'
    ;
```
In TypeScript, I want to avoid numeric enums wherever possible, and avoid having to prefix values with a type (e.g. DialogIcon.Info).

packages/sdk/src/types/runtime/dialog.ts
```diff
@@ -0,0 +1,86 @@
...
+		text: string,
+		recipients?: GroupMask | User[] | User,
+		icon?: DialogIcon | Texture,
+		persistent?: boolean
+	}) {
+		super();
+	}
+
+	/** Convenience method for broadcasting a message */
+	public static Message(context: Context, text: string): Dialog { }
```
eanders-MS: I think the fluent API mentioned above could achieve the same goal as these convenience methods, but with more flexibility. Your thoughts?

dialog_design.md
```diff
@@ -0,0 +1,68 @@
...
+	1. A `message` dialog presents an OK button, and returns no value, just a confirmation that OK was clicked.
+	2. A `confirm` dialog presents OK and Cancel buttons, and returns a boolean.
+	3. A `prompt` dialog presents a text field and OK/Cancel buttons, and returns the contents of the text field.
+2. `text` - The text that the dialog displays.
+3. `recipients` (optional) - The dialog is displayed to all currently connected users by default, but can be restricted
+	to a subset of users with this option. Takes either a `GroupMask`, a `User[]`, or just one `User`. This filtering
+	could be implemented in the sync layer fairly easily, I think.
+4. `icon` (optional) - A graphic to display alongside the text. Can be one of the built-in icons (`info`, `warning`,
+	`error`, or `question`), or a custom texture supplied by the developer.
+5. `persistent` (optional) - If `true`, the dialog is presented to users in the recipients list even if they are
```
eanders-MS: Let's defer `persistent` in v1. Sounds tricky, and there's enough value here without the complication.

stevenvergenz:
I wonder if a dialog system is overkill. With the exception of the text input field, everything else could be implemented better as a node library. Is there a different way we could integrate text input?

dialog_design.md
```diff
@@ -0,0 +1,68 @@
...
+Goal: Display a message and an optional graphic to all users, or some, or one, and return a response.
+
+## Architecture
+
+Create a new `Dialog` class that manages calls and responses. It is also a good place to park static convenience
+functions and event listeners. I propose a couple different setup options:
+
+1. `type` - What sort of response we should expect.
+	1. A `message` dialog presents an OK button, and returns no value, just a confirmation that OK was clicked.
+	2. A `confirm` dialog presents OK and Cancel buttons, and returns a boolean.
```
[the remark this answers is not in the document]
eanders-MS: I like this. Perhaps an array of buttons: ['ok', 'cancel'].

dialog_design.md
eanders-MS: It would be nice if `show()` returned a `Promise<DialogResult[]>` (see mention of the `DialogResult` type below).

```ts
    const results = await Dialog.Create(this.context, options).show();
```

dialog_design.md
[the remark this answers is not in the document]
sorenhan: Yeah, if we make this an array of buttons, 'message' and 'confirm' are really the same thing. then the type field really comes down to 'add a text field'. 
We would need to also add a "default result" for if you just press enter on a text field

---
In reply to: [320502157](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/444#discussion_r320502157) [](ancestors = 320502157)

dialog_design.md
sorenhan: Returning a promise works if there was just one user getting the result but with multiple recipients, would that work?

---
In reply to: [320870236](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/444#discussion_r320870236) [](ancestors = 320870236)

dialog_design.md
```diff
@@ -0,0 +1,68 @@
+# Dialogs
+
+Goal: Display a message and an optional graphic to all users, or some, or one, and return a response.
+
+## Architecture
+
+Create a new `Dialog` class that manages calls and responses. It is also a good place to park static convenience
```
sorenhan: >Create a new `Dialog` class that manages calls and responses [](start = 0, length = 60)

I think it should an actor, like text (in the future a component).
 a. solves ownership, destruction, transform.
 b. solves user access - single-user/groups/everyone
 c. we could add an option to close on selection (default to on) - leave the dialog up after clicks, or just auto-close.
 d. we could add an option to locally user-face it (that's one reason why it's a child of an actor, not an actor - it's per-user only)
 e. solves your "persistent" suggestion - it's automatically synced to new users that meet the filtering requirements.
  f. could have option to just re-show the dialog.

sorenhan:
This initially started because we wanted a text input field, so we could argue that.
There are at least 3 additional benefits to have a dialog system 
1. no latency on input selection.
2. allows custom dialog UI per host app, or per engine integration
3. enables per engine optimizations that generic actors wouldn't

---
In reply to: [527662800](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/444#issuecomment-527662800) [](ancestors = 527662800)

dialog_design.md
stevenvergenz: It seems like a stretch to call something with this functionality an Actor. There's no transform (it's modal), there's no hierarchy, and there's no global state to synchronize. None of the other Actor components would make sense on a dialog (collider, rigidbody, mesh, light...). If anything, this is the first piece of our UI infrastructure, but even that is a stretch since it's unsynchronized by necessity (how weird would it be if you could see other people type into your modal dialog?)

---
In reply to: [321512104](https://github.com/microsoft/mixed-reality-extension-sdk/pull/444#discussion_r321512104) [](ancestors = 321512104)

dialog_design.md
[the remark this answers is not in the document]
stevenvergenz: That's not bad. Instead of static ok/cancel, take an array of strings and we'll have that many buttons. It would respond with the text string on the clicked button.

---
In reply to: [321510661](https://github.com/microsoft/mixed-reality-extension-sdk/pull/444#discussion_r321510661) [](ancestors = 321510661,320502157)

dialog_design.md
stevenvergenz: Agreed, promise doesn't really work if there are multiple responses. I guess it could get wired up to the `finished` event though.

---
In reply to: [321511608](https://github.com/microsoft/mixed-reality-extension-sdk/pull/444#discussion_r321511608) [](ancestors = 321511608,320870236)

packages/sdk/src/types/runtime/dialog.ts
stevenvergenz: I agree with avoiding numeric enums, but why no type prefixes? We do that with primitive types, and I think it aids in discoverability.

---
In reply to: [320499134](https://github.com/microsoft/mixed-reality-extension-sdk/pull/444#discussion_r320499134) [](ancestors = 320499134)

packages/sdk/src/types/runtime/dialog.ts
stevenvergenz: I'd rather have an initializer object, there's more precedent for it. In terms on in-lining handlers, you can already do that:
```ts
const dialog = new Dialog(this.context, {
    type: 'message',
    icon: 'info',
    text: "You lost the game",
    recipients: loser
})
.on('response', () => console.log('User acknowelged'))
.show();
```

---
In reply to: [320498358](https://github.com/microsoft/mixed-reality-extension-sdk/pull/444#discussion_r320498358) [](ancestors = 320498358)

packages/sdk/src/types/runtime/dialog.ts
stevenvergenz: Versus `(new Dialog(context, options)).show()`? Maybe, but it might confuse people that they're interchangeable.

---
In reply to: [320858387](https://github.com/microsoft/mixed-reality-extension-sdk/pull/444#discussion_r320858387) [](ancestors = 320858387)

packages/sdk/src/types/runtime/dialog.ts
```diff
@@ -0,0 +1,86 @@
...
+	public static PromptPrivate(context: Context, user: User, text: string): Promise<string> { }
+
+	/** Broadcast the dialog */
+	public send() { }
+
+	/** Close the dialog on all clients, and clear persistence. */
+	public cancel() { }
+
+	/** Fires every time a user responds to the dialog. */
+	public on(eventType: 'response', listener: (user: User, response: string | boolean) => void): this;
```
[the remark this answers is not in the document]
stevenvergenz: I like this idea. Though `response` is referring to the user response, not the network response. And in the failure case, I think I'd rather leave result empty and have a separate error field. Or better still, a whole separate event for errors. Counting errors as a response is more network-first thinking.

---
In reply to: [320869055](https://github.com/microsoft/mixed-reality-extension-sdk/pull/444#discussion_r320869055) [](ancestors = 320869055)

packages/sdk/src/types/runtime/dialog.ts
eanders-MS: The main benefit of this format over something prefixed is that it is directly representable in JSON, and this format seems to be where industry is going. For example, see the DOM definition file `C:\Users\<alias>\AppData\Local\Programs\Microsoft VS Code\resources\app\extensions\node_modules\typescript\lib\lib.dom.d.ts`. Toward the end of the file there are over 100 of these kinds of strongly-typed string unions defined.

packages/sdk/src/types/runtime/dialog.ts
eanders-MS: You have an unused var (`dialog`) in your example. This would trigger a lint error. You could work around it with parens, but that's just fighting the linter. A static `Create` method would solve it more cleanly. I had a comment to this effect in an earlier iteration, but don't see it now. Weird.

packages/sdk/src/types/runtime/dialog.ts
eanders-MS: I think discoverability is good with string union types. Developers are used to them.

packages/sdk/src/types/runtime/dialog.ts
```diff
@@ -0,0 +1,65 @@
...
+	 * @param options.text The message to present.
+	 * @param options.buttons Label strings for dialog buttons. The user response will contain this string.
+	 * @param options.input Whether the dialog should present a text input. Defaults to false.
+	 * @param options.recipients Who to send the message to. Defaults to all users.
+	 * @param options.icon A graphic to display next to the text. Defaults to none.
+	 * @param options.persistent Send this message to late-joining users. Defaults to false.
+	 */
+	public constructor(private context: Context, options: {
+		text: string,
+		buttons: string[],
```
eanders-MS: Would we enforce a maximum of two button labels?

packages/sdk/src/types/network/payloads/payloads.ts
```diff
@@ -354,3 +356,29 @@ export type InterpolateActor = Payload & {
...
+
+/**
+ * @hidden
+ * Engine to app. Acknowledgement of modal dialog.
+ */
+export type DialogResponse = Payload & {
+	type: 'dialog-response';
+	dialogId: string;
+	userId: string;
+	button: boolean;
```
eanders-MS: Update to `button: string;`

dialog_design.md
eanders-MS: I think it would be wrong to make this an Actor component. This is local UI. It should never be synced or visible to other players.
- The dialog shouldn't have a transform. It should be view-locked.
- single-user/groups/everyone is already solved in this design.
- when new users join, the app should decide if they should see the dialog and can show it to them. This would be straight forward to do in the userJoined callback.

Soren did you have a use case in mind where an Actor-based dialog was a better solution?

dialog_design.md
```diff
@@ -0,0 +1,72 @@
...
+public on(eventType: 'response', listener: (r: DialogResponse) => void): this;
+```
+
+And another when all the responses are in, if it's not a persistent dialog. This version requires collecting all the
+responses inside the dialog class so they can be passed together into this callback:
+
+```ts
+public on(eventType: 'finished', listener: (responses: DialogResponse[]) => void): this;
+```
+
```
sorenhan: I don't think "Finished" makes sense. The concept of "all responses" seems strange to me - you can't ever expect all users to interact with an app. If just a single user doesn't interact then this won't trigger, so this would rarely be useful

packages/sdk/src/types/runtime/dialog.ts
```diff
@@ -0,0 +1,86 @@
...
+	/**
+	 * Create a new dialog box to present to users.
+	 * @param context An active MRE context.
+	 * @param options.type How to present the choice to users.
+	 * @param options.text The message to present.
+	 * @param options.recipients Who to send the message to. Defaults to all users.
+	 * @param options.icon A graphic to display next to the text. Defaults to none.
+	 * @param options.persistent Send this message to late-joining users. Defaults to false.
+	 */
+	public constructor(private context: Context, options: {
```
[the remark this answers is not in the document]
stevenvergenz: What would a dialog without text or buttons even do? I don't think you can get away without specifying *something*.

packages/sdk/src/types/runtime/dialog.ts
stevenvergenz: I can think of three-button use cases (retry, ignore, cancel). Are there use cases that need more, say 10? If it's easy for us to reflow the button panel, I see no need to arbitrarily limit it.

dialog_design.md
stevenvergenz: It makes sense without the `persistent` property. You're choosing a list of users to present the dialog to, so it makes sense to expect them to be active participants. Or else you're sending it to a single user on `onUserJoined`, in which case it's interchangeable with `onResponse`.

tombuMS:
Looks good to me

Issue #374, opened 2019-07-08
Add SDK performance counters


eanders-MS:

sorenhan:
Maybe loaded asset counts (number of actors, meshes, polys, materials, textures, pixels) is the highest priority.
Maybe bandwidth used?

sorenhan:
and number of actors

stevenvergenz:
My proposal for stats to track:
* Actor count
* Prefab count
* Sound count
* Material count
* Texture count
    * Total texture pixels
* Mesh count
    * Total mesh polygons
    * Total mesh vertices
* Average websocket bandwidth in/out over the last 5/30/60 seconds
* Network message count in/out over the last 5/30/60 seconds

sorenhan:
I would also add 
1. "sound time in seconds" - as that's a pretty good measure of how much memory the sound waveforms take up,
2. Active Media Instances. - how many sounds/video streams are currently playing.

PR #445, opened 2019-09-09
Add performance stats API

stevenvergenz:
Implements #374

### DIFF
--- a/packages/sdk/src/adapters/multipeer/session.ts
+++ b/packages/sdk/src/adapters/multipeer/session.ts
@@ -9,7 +9,7 @@ import {
 	Client, InitializeActorMessage, MissingRule, Rules, SessionExecution,
 	SessionHandshake, SessionSync, SyncActor, SyncAsset
 } from '.';
-import { Connection, Message, UserLike } from '../..';
+import { Connection, EventedConnection, Message, UserLike } from '../..';
 import { log } from '../../log';
 import * as Protocols from '../../protocols';
 import * as Payloads from '../../types/network/payloads';
@@ -166,10 +166,24 @@ export class Session extends EventEmitter {
 			// tslint:disable-next-line:no-console
 			log.error('network', `[ERROR] setAuthoritativeClient: client ${clientId} does not exist.`);
 		}
-		this._clientSet[clientId].setAuthoritative(true);
-		this.clients
-			.filter(client => client.id !== clientId && client.isJoined())
-			.forEach(client => client.setAuthoritative(false));
+		const oldAuthority = this.authoritativeClient;
+		const newAuthority = this._clientSet[clientId];
+
+		newAuthority.setAuthoritative(true);
+		oldAuthority.setAuthoritative(false);
+
+		// forward network stats from the authoritative peer connection to the app
+		const toApp = this.conn instanceof EventedConnection ? this.conn : null;
+		const toOldAuthority = oldAuthority.conn instanceof EventedConnection ? oldAuthority.conn : null;
+		const toNewAuthority = newAuthority.conn instanceof EventedConnection ? newAuthority.conn : null;
+		if (toApp && toOldAuthority && toNewAuthority) {
+			const forwardIncoming = (bytes: number) => toApp.statsTracker.recordIncoming(bytes);
+			const forwardOutgoing = (bytes: number) => toApp.statsTracker.recordOutgoing(bytes);
+			toOldAuthority.statsTracker.off('incoming', forwardIncoming);
+			toOldAuthority.statsTracker.off('outgoing', forwardOutgoing);
+			toNewAuthority.statsTracker.on('incoming', forwardIncoming);
+			toNewAuthority.statsTracker.on('outgoing', forwardOutgoing);
+		}
 	}
 
 	private recvFromClient = (client: Client, message: Message) => {
--- a/packages/sdk/src/adapters/multipeer/session.ts
+++ b/packages/sdk/src/adapters/multipeer/session.ts
@@ -170,20 +170,24 @@ export class Session extends EventEmitter {
 		const newAuthority = this._clientSet[clientId];
 
 		newAuthority.setAuthoritative(true);
+		for (const client of this.clients.filter(c => c !== newAuthority)) {
+			client.setAuthoritative(false);
+		}
 
 		// forward network stats from the authoritative peer connection to the app
 		const toApp = this.conn instanceof EventedConnection ? this.conn : null;
-		const toNewAuthority = newAuthority.conn instanceof EventedConnection ? newAuthority.conn : null;
 		const forwardIncoming = (bytes: number) => toApp.statsTracker.recordIncoming(bytes);
 		const forwardOutgoing = (bytes: number) => toApp.statsTracker.recordOutgoing(bytes);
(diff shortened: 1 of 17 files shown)


packages/sdk/src/protocols/execution.ts
```diff
@@ -30,10 +30,15 @@ import { Sync } from './sync';
  * Class to handle operational messages with a client.
  */
 export class Execution extends Protocol {
+	public networkStats: NetworkStats;
```
eanders-MS: This should go on the `Connection` class, as a peer of `ConnectionQuality`.

packages/sdk/src/protocols/networkStats.ts
```diff
@@ -0,0 +1,101 @@
...
+
+	constructor() {
+		this.beforeSend = this.beforeSend.bind(this);
+		this.beforeRecv = this.beforeRecv.bind(this);
+		setInterval(() => this.cycleStatFrames(), 1000);
+	}
+
+	/** @private */
+	public beforeSend(message: Message, promise?: ExportedPromise): Message {
+		const data = new Buffer(JSON.stringify(message, (key, value) => {
```
eanders-MS: Benefit to putting `NetworkStats` on `Connection`: You won't have to double-serialize the message to get the size.

packages/sdk/src/protocols/networkStats.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { Middleware } from '.';
+import { Message } from '..';
+import { ExportedPromise } from '../utils/exportedPromise';
+import filterEmpty from '../utils/filterEmpty';
+import validateJsonFieldName from '../utils/validateJsonFieldName';
+
+export interface NetworkStatsFrame {
```
eanders-MS: >interface [](start = 7, length = 9)

For consistency, prefer `type` over `interface`. Reserve `interface` for types that will be implemented by a class.

packages/sdk/src/protocols/networkStats.ts
stevenvergenz: Actually that's not true. The Pipe connection class deals with objects and not serialized data, so it would still have to manually serialize.

packages/sdk/src/adapters/multipeer/session.ts
```diff
@@ -166,10 +166,27 @@ export class Session extends EventEmitter {
 			// tslint:disable-next-line:no-console
 			log.error('network', `[ERROR] setAuthoritativeClient: client ${clientId} does not exist.`);
 		}
-		this._clientSet[clientId].setAuthoritative(true);
-		this.clients
-			.filter(client => client.id !== clientId && client.isJoined())
-			.forEach(client => client.setAuthoritative(false));
+		const oldAuthority = this.authoritativeClient;
+		const newAuthority = this._clientSet[clientId];
```
eanders-MS: I was hoping this could be handled on pipe.remote. Getting the stats from the authoritative peer is a lot of new code.

packages/sdk/src/adapters/multipeer/session.ts
```diff
@@ -166,10 +166,27 @@ export class Session extends EventEmitter {
...
+		const forwardIncoming = (bytes: number) => toApp.statsTracker.recordIncoming(bytes);
+		const forwardOutgoing = (bytes: number) => toApp.statsTracker.recordOutgoing(bytes);
+		toNewAuthority.statsTracker.on('incoming', forwardIncoming);
+		toNewAuthority.statsTracker.on('outgoing', forwardOutgoing);
+
+		// turn off old authority
+		const toOldAuthority = oldAuthority && oldAuthority.conn instanceof EventedConnection
+			? oldAuthority.conn : null;
+		if (oldAuthority) {
+			oldAuthority.setAuthoritative(false);
```
eanders-MS: You changed the behavior here. We used to send .setAuthoritative(false) to all other clients. You might be introducing a bug here (or exposing one). Safer to keep it the way it was.

packages/sdk/src/protocols/networkStats.ts
eanders-MS: oh you're right. that is more efficient, but annoying in this instance.

---
In reply to: [325397097](https://github.com/microsoft/mixed-reality-extension-sdk/pull/445#discussion_r325397097) [](ancestors = 325397097)

packages/sdk/src/connection/networkStats.ts
```diff
@@ -0,0 +1,90 @@
...
+
+/** @hidden */
+type NetworkStatsFrame = {
+	messageCount: number;
+	trafficIn: number;
+	trafficOut: number;
+};
+
+/** A collection of network statistics from a certain point in time. */
+export interface NetworkStatsReport {
```
eanders-MS: >interface [](start = 7, length = 9)

This should be a `type`, not an `interface`.

packages/sdk/src/types/performanceStats.ts
```diff
@@ -0,0 +1,71 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * A snapshot of various stats useful for debugging and benchmarking. None of these correlate exactly with
+ * a particular client's experience, framerate, memory usage, etc., but should be generally indicative.
+ */
+export interface PerformanceStats {
```
eanders-MS: >interface [](start = 7, length = 9)

Should be a `type` not an `interface`

packages/sdk/src/adapters/multipeer/session.ts
stevenvergenz: Nothing gets serialized or deserialized anywhere but in a WebSocket connection. If we did it in the pipe it would add overhead.

---
In reply to: [325909690](https://github.com/microsoft/mixed-reality-extension-sdk/pull/445#discussion_r325909690) [](ancestors = 325909690)

eanders-MS: Agreed this is better approach than adding overhead in in pipe.remote.

---
In reply to: [325918122](https://github.com/microsoft/mixed-reality-extension-sdk/pull/445#discussion_r325918122) [](ancestors = 325918122,325909690)

packages/sdk/src/types/performanceStats.ts
```diff
@@ -0,0 +1,68 @@
...
+	 * to the bandwidth sent to the busiest client over the interval, though this does not correlate exactly.
+	 * Only MRE internal traffic is counted, not general HTTP requests (static file hosting, etc.).
+	 */
+	networkBandwidthOut: [number, number, number];
+	/**
+	 * The number of messages sent and received by this app in the last 1/5/30 seconds. A high number might indicate
+	 * that clients are wasting CPU cycles serializing and deserializing messages.
+	 */
+	networkMessageCount: [number, number, number];
+}
```
eanders-MS: Needs a semicolon

eanders-MS:
> 

Needs a semicolon (or tslint will complain)

---
Refers to: packages/sdk/src/connection/networkStats.ts:91 in d857e57. [](commit_id = d857e57f97244900ecc1c9e593d5f09af0be512b, deletion_comment = False)

PR #453, opened 2019-10-07
WebHost: add API for custom connection handlers

eanders-MS:
#### `WebHost.addWebSocketRoute`

This is a feature needed by the WebProjector. Added an API to WebHost to allow the registration of custom WebSocket connection handlers at different routes, e.g.:
```ts
webHost.addWebSocketRoute("/my/custom/endpoint", myCustomConnectionHandler);
```
The first parameter is the route, and can be a string or a RegExp. Second parameter is the connection handler.

#### `WebHost.addAdapter`

Building on the above capability, it was trivial to add support for servicing multiple MREs from a single WebHost, so I went ahead and added that:
```ts
const functionalTestAdapter = new MultipeerAdapter();
const helloWorldAdapter = new MultipeerAdapter();
functionalTestAdapter.onConnection((context, params) => new FunctionalTestApp(context, params));
helloWorldAdapter.onConnection((context, params) => new HelloWorldApp(context, params));
webHost.addAdapter("/functional-tests", functionalTestAdapter);
webHost.addAdapter("/hello-world", helloWorldAdapter);
```

All changes are backward compatible with existing code.

**UPDATE**: Hiding `WebHost.addAdapter` for now. There are design complications with it and serving static buffers.

### DIFF
--- a/packages/sdk/src/adapters/websocket/adapter.ts
+++ b/packages/sdk/src/adapters/websocket/adapter.ts
@@ -5,22 +5,15 @@
 
 import * as http from 'http';
 import QueryString from 'query-string';
-import * as Restify from 'restify';
 import UUID from 'uuid/v4';
 import * as WS from 'ws';
-import { Adapter, AdapterOptions } from '..';
+import { Adapter } from '..';
 import { Context, WebSocket } from '../..';
 import * as Constants from '../../constants';
-import verifyClient from '../../utils/verifyClient';
 import { log } from './../../log';
 
 // tslint:disable-next-line:no-var-requires
 const forwarded = require('forwarded-for');
-
-/**
- * WebSocket Adapter options.
- */
-export type WebSocketAdapterOptions = AdapterOptions;
 
 /**
  * The `WebSocketAdapter` is appropriate to use when the host environment has an authoritative simluation, and that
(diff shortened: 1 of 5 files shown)


stevenvergenz:
It's cool that multiplexing apps is possible now, but I still doubt whether it's desirable from a performance perspective.

eanders-MS:
Depends a lot on the apps involved, but could be a nice money saver for hobbyists.

---
In reply to: [539228896](https://github.com/microsoft/mixed-reality-extension-sdk/pull/453#issuecomment-539228896) [](ancestors = 539228896)

stevenvergenz:
I suppose that's true, it depends on how their host structures it. On Azure you only pay for the app service plan, which can host multiple apps.

---
In reply to: [539241977](https://github.com/microsoft/mixed-reality-extension-sdk/pull/453#issuecomment-539241977) [](ancestors = 539241977,539228896)

packages/sdk/src/webHost.ts
```diff
@@ -3,82 +3,148 @@
...
+			log.info('app', `baseUrl: ${this.baseUrl}`);
+			log.info('app', `baseDir: ${this.baseDir}`);
+			if (!!this.baseDir) {
+				serveStaticFiles();
 			}
-			));
+			handleWebSocketUpgrades();
+		});
+
+		const serveStaticFiles = () => {
```
stevenvergenz: Static file hosting should be served from the adapter host as well. Two apps in the same server aren't going to necessarily merge their public directories.

packages/sdk/src/adapters/websocket/adapter.ts
```diff
@@ -31,69 +24,40 @@ export type WebSocketAdapterOptions = AdapterOptions;
...
-			this.startListening();
-			return Promise.resolve(this.server);
-		}
-	}
-
-	private startListening() {
-		// Create a server for upgrading HTTP connections to WebSockets.
-		const wss = new WS.Server({ server: this.server, verifyClient });
+	/** @inheritdoc */
+	public get name(): string { return this. constructor.name; }
```
stevenvergenz: Is there a space here? Does that compile?

packages/sdk/src/webHost.ts
```diff
@@ -3,82 +3,148 @@
...
-				log.info('app', `baseUrl: ${this.baseUrl}`);
-				log.info('app', `baseDir: ${this.baseDir}`);
-				if (!!this.baseDir) {
-					this.serveStaticFiles(server);
-				}
-			})
-			.catch(reason => log.error('app', `Failed to start HTTP server: ${reason}`));
-	}
+		// If an adapter was provided, register it.
+		if (this.options.adapter) {
```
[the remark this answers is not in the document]
eanders-MS: In fact the value can be `null` here. The test above is a strict check for `undefined`, where this is a softer "falsey" check. This allows the user to pass `null` to interrupt the default behavior of allocating a MultipeerAdapter internally. I did it this way for backward compatibility with existing code.

---
In reply to: [332694452](https://github.com/microsoft/mixed-reality-extension-sdk/pull/453#discussion_r332694452) [](ancestors = 332694452)

packages/sdk/src/webHost.ts
eanders-MS: Interesting. For the sake of backward compatibility I'd like to keep it this way for now. Alternately, I can remove support for multiple MREs. It seemed like a neat thing to enable, but this makes it complicated. I don't think WebHost is the right solution for large-scale apps anyway. We shouldn't presume to own the http server or WebSocket server.

---
In reply to: [332689431](https://github.com/microsoft/mixed-reality-extension-sdk/pull/453#discussion_r332689431) [](ancestors = 332689431)

packages/sdk/src/adapters/websocket/adapter.ts
eanders-MS: Weird. Yep it compiles. Will fix.

---
In reply to: [332696471](https://github.com/microsoft/mixed-reality-extension-sdk/pull/453#discussion_r332696471) [](ancestors = 332696471)

packages/sdk/src/webHost.ts
stevenvergenz: I'm in favor of removing multi-MRE support if we can't do it well, or else document the interrelationships.

---
In reply to: [332711372](https://github.com/microsoft/mixed-reality-extension-sdk/pull/453#discussion_r332711372) [](ancestors = 332711372,332689431)

eanders-MS: Multi-MRE support disabled for now.

PR #454, opened 2019-10-08
Reinstate altspacevr-extras

eanders-MS:
Also updated it to the latest RPC interface.


stevenvergenz:
Please add the line to copy this package back into the Dockerfile.

PR #455, opened 2019-10-11
[SDK] Collision layers

stevenvergenz:
* Add design doc for collision layers.
* Implement the (minimal) changes needed for layers in the SDK.
* Add collision layer functional test.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -25,7 +25,10 @@
       "internalConsoleOptions": "openOnSessionStart",
       "autoAttachChildProcesses": true,
       "console": "internalConsole",
-      "outputCapture": "std"
+	  "outputCapture": "std",
+	  "env": {
+		  // "MRE_LOGGING": "network"
+	  }
     }
   ]
 }
--- a/design/collider.layer.md
+++ b/design/collider.layer.md
@@ -1,0 +1,44 @@
+Collision Layers
+=================
+
+Collision layers are hard, because every engine does it differently. I will attempt to document how it works in each major physics engine here:
+
+* **Unity (static one:many)** - Each object in the scene is assigned exactly one layer (not inherited). The matrix of which layers collide with which other layers is fixed at compile time (configurable from Project Settings). In addition to being used for collisions, these layers are also used by cameras to limit what is rendered, by the lighting system to determine which lights shine on an object, and which objects are hit by a raycast. The global maximum layer count is 31.
+
+* **Unreal 4 (dynamic one:many)** - Each object in the scene is assigned exactly one "object type" and a list of how that object responds to every other object type. All these can be set at runtime. Raycasts are also assigned a type from the list, so the interaction scheme is the same. The global maximum type count is 32.
+
+* **Bullet (dynamic many:many)** - Each object is assigned a "collision groups" bitfield32 and a "collision mask" bitfield32. Two objects only collide if one of object A's groups is in object B's mask, and vice versa. Raycasts work the same way, with group and mask fields.
+
+With this in mind, I propose the following architecture:
+
+API Design
+-----------
+
+Add a field to the `Collider` and `ColliderLike` types: `layer: CollisionLayer`.
+
+Layers
+-------
+
+I believe a fixed collision layer strategy can cover the most common use cases. In this proposal, there are four collision layers:
+
+```ts
+type CollisionLayer
+	= 'default'
+	| 'ui'
+	| 'navigation'
+	| 'hologram';
+```
+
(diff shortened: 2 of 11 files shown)


packages/sdk/src/types/runtime/collider.ts
```diff
@@ -8,13 +8,28 @@ import { InternalCollider } from '../internal/collider';
...
+/**
+ * Controls what the assigned actors will collide with.
+ * * `default` - Good for most actors. These will collide with all "physical" things: other default actors,
+ * navigation actors, and the non-MRE environment. It also blocks the UI cursor and receives press/grab events.
+ * * `navigation` - For actors considered part of the environment. Can move/teleport onto these colliders,
+ * but cannot click or grab them.
+ * * `hologram` - For "non-physical" actors. Only interact with the cursor and other holograms.
+ * * `ui` - Actors in this layer do not collide with anything but the UI cursor.
+ */
+export type CollisionLayer
```
tombuMS: I though that we were moving away from the type model in favor of enums for this sort of thing.  Is that not the case any longer?

design/collider.layer.md
```diff
@@ -0,0 +1,58 @@
...
+* `navigation` - For actors considered part of the environment. Can move/teleport onto these colliders, but cannot click or grab them.
+* `hologram` - For "non-physical" actors. Only interact with the cursor and other holograms.
+* `ui` - Actors in this layer do not collide with anything but the UI cursor.
+
+| Self \ Other  | Default | Navigation | Hologram | UI | Cursor | Environment | Player |
+|---------------|---------|------------|----------|----|--------|-------------|--------|
+| Default       | X       | X          |          |    | X      | X           |        |
+| Navigation    | X       | X          |          |    | *      | X           | X      |
+| Hologram      |         |            | X        |    | X      |             |        |
+| UI            |         |            |          |    | X      |             |        |
```
tombuMS: What would attachments on the player be considered in this layering paradigm?

packages/sdk/src/types/runtime/collider.ts
stevenvergenz: Good catch. I was remembering feedback I got on a previous PR to change to a type, but that was an interface, not an enum.

design/collider.layer.md
stevenvergenz: Those actors can still be anything, there's no restriction in the SDK. Though maybe it would make sense to disallow navigation actors to be attached to the player, to prevent surfing.

eanders-MS:
Actually type is preferred over enum. Want to rid the codebase of enums.

Sent from 📱
________________________________

packages/sdk/src/types/runtime/collider.ts
stevenvergenz: I prefer enums personally. The primary reason for this preference is the ease of documentation. Individual values of a type cannot be documented with typedoc, but enum values can.

eanders-MS: To close the loop... Steven and I discussed the merits of `type` vs `enum` for use-cases like this one, and enums appear to be the better choice. I was swayed primarily by the fact that it isn't possible to add docstrings to individual elements of a union type (well, you can but it's unwieldy).

If no objection, going forward let's standardize on using "enum of string literals" for enum-like fields rather than "typed union of string literals".

I'll file an issue to track the work of fixing up the rest of the codebase. There will be some edge cases.

eanders-MS:
Layer is inherited, correct?

eanders-MS:
Answer: not inherited.

---
In reply to: [541977403](https://github.com/microsoft/mixed-reality-extension-sdk/pull/455#issuecomment-541977403) [](ancestors = 541977403)

.vscode/launch.json
```diff
@@ -25,7 +25,10 @@
       "internalConsoleOptions": "openOnSessionStart",
       "autoAttachChildProcesses": true,
       "console": "internalConsole",
-      "outputCapture": "std"
+	  "outputCapture": "std",
+	  "env": {
+		  // "MRE_LOGGING": "network"
```
eanders-MS: >// "MRE_LOGGING": "network" [](start = 4, length = 27)

Comments are not valid in JSON, but maybe it's fine in vscode.

stevenvergenz: VS Code didn't complain. I find myself putting this in there fairly regularly, it'll save us time next time to have this here.

---
In reply to: [335211607](https://github.com/microsoft/mixed-reality-extension-sdk/pull/455#discussion_r335211607) [](ancestors = 335211607)

eanders-MS: 👍

PR #463, opened 2019-10-31
[minor] Make PrimitiveDefinition a discriminated union

eanders-MS:
It was bothering me that invalid fields like `uSegments` were showing up in intellisense when creating a box primitive, so I modified the type definition to discriminate by primitive type.

At the same time, I changed `ColliderType` to be an enum, to align with our newly established coding guideline that we don't use string-union types where enums would suffice.

TODO: ColliderType needs docstrings.
TODO: *PrimitiveDefinition needs better docstrings.

### DIFF
--- a/packages/sdk/src/types/primitiveTypes.ts
+++ b/packages/sdk/src/types/primitiveTypes.ts
@@ -16,28 +16,78 @@ export enum PrimitiveShape {
 	Plane = 'plane'
 }
 
-/**
- * The size, shape, and description of a primitive.
- */
-export interface PrimitiveDefinition {
-	// TODO: Make this a discriminated union type.
-	/**
-	 * The general shape of the defined primitive.
-	 */
-	shape: PrimitiveShape;
-
+export type SpherePrimitiveDefinition = {
+	shape: PrimitiveShape.Sphere
 	/**
 	 * The bounding box size of the primitive.
 	 */
 	dimensions?: Partial<Vector3Like>;
-
 	/**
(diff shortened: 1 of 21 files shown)


packages/sdk/src/types/primitiveTypes.ts
```diff
@@ -16,28 +16,78 @@ export enum PrimitiveShape {
...
+	 */
+	uSegments?: number;
 	/**
 	 * The number of vertical or axial segments of spheres, capsules, and planes.
 	 */
 	vSegments?: number;
-}
+};
+
+export type PrimitiveDefinition
```
stevenvergenz: This is not a very friendly data structure, which is why the assets API doesn't use it. I would be in favor of leaving this unchanged as a network type, then having a separate discriminated union for the CreatePrimitive API. That way we could have a `radius` field for example, but still have a unified data structure for the network message.

eanders-MS: I like that approach. I was wondering why we didn't have a radius field.

eanders-MS:
@stevenvergenz old apps that use primitives will still compile, but old apps that use colliders won't due to the change to `ColliderType` from union to enum, but I think it's a good change and worth it.

This change was just meant to clean up our API a little bit (ColliderType change) and improve intellisense (PrimitiveDefinition change). I like your suggestion above about redefining the PrimitiveDefinition types to be more friendly, but that's beyond scope for this.

No network payloads were changed, so it's backward compatible from that perspective.

PR #464, opened 2019-11-01
[RFC] New animation system proposal

stevenvergenz:


design/animation.md
```diff
@@ -0,0 +1,317 @@
...
+For many purposes, creating and managing an animation is overkill. This is why objects that can be animated have a
+convenient fire-and-forget function: `animateTo`. This function takes a "destination" ActorLike or MaterialLike and
+a duration as input, generates a simple one-frame animation from all the fields given in the -Like, binds it to the
+object, and plays it.
+
+```ts
+class Animation {
+	public static AnimateTo(target: Animatible, options: {
+		duration: number;
+		destination: ActorLike | MaterialLike;
```
eanders-MS: Should this be `target`?

eanders-MS: I think this needs to be `Partial<ActorLike> | Partial<MaterialLike>`, so that users don't have to pass a fully filled out value. It will make statically validating them impossible, but would alleviate the need for users to cast them, e.g. `as MaterialLike`.

design/animation.md
```diff
@@ -0,0 +1,317 @@
...
+	/** Current play state */
+	isPlaying: Readonly<boolean>;
+
+	play(): void;
+	pause(): void;
+
+	/** Only fired if wrapMode is Once */
+	finished(): Promise<void>;
+	/** Sets the `continueWith` property, and returns the next anim instance */
+	continueWith(otherAnim: AnimationInstace): AnimationInstance;
```
eanders-MS: `continueWith` is defined twice in this interface.

design/animation.md
```diff
@@ -0,0 +1,317 @@
...
+type Keyframe<T extends AnimationProp> = {
+	time: number;
+	value: T;
+	easing?: EasingFunction;
+}
+
+type EasingFunction = [number, number, number, number];
+```
+
+### Paths
```
eanders-MS: Can you add an example of interpolating one actor to another actor's position?

eanders-MS: I'd like to see the syntax for referencing the field of another actor as a target value.

design/animation.md
eanders-MS: `continueWith` is specialized functionality that should live outside this class. I'd put this functionality into a new class called `AnimationGroup`, or `AnimationChain`. It would consist of a set of `AnimationInstance`s and rules for how to play them together. Like a command buffer for animations.

design/animation.md
```diff
@@ -0,0 +1,317 @@
...
+	id: Readonly<string>;
+	duration: Readonly<number>;
+	targetCount: Readonly<number>;
+}
+
+class AssetContainer {
+	public createAnimation(name: string, options: { targetCount?: number, tracks: Track<AnimationProp>[] }): Animation;
+}
+
+type AnimationProp = Vector2 | Vector3 | Quaternion | Color3 | Color4 | number | string | boolean;
```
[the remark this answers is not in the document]
stevenvergenz: I'm flexible, either way is fine. And there are probably more types we can add to this list during implementation. This is just a minimum set I think.

---
In reply to: [341682310](https://github.com/microsoft/mixed-reality-extension-sdk/pull/464#discussion_r341682310) [](ancestors = 341682310)

design/animation.md
stevenvergenz: I thought a linked-list type chain like this would be simpler and easier to understand. I'm curious about what you have in mind though, any more details?

---
In reply to: [341702219](https://github.com/microsoft/mixed-reality-extension-sdk/pull/464#discussion_r341702219) [](ancestors = 341702219)

design/animation.md
stevenvergenz: Whoops, I totally overlooked that use case. Will update.

---
In reply to: [341698633](https://github.com/microsoft/mixed-reality-extension-sdk/pull/464#discussion_r341698633) [](ancestors = 341698633)

design/animation.md
stevenvergenz: I didn't want to confuse this option with the other `target` parameter. Am open to naming changes.

---
In reply to: [341689594](https://github.com/microsoft/mixed-reality-extension-sdk/pull/464#discussion_r341689594) [](ancestors = 341689594)

design/animation.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+Fade out a material over one second:
+
+```ts
+const anim: MRE.Animation = material.animateTo({
+	duration: 1,
+	easing: MRE.Animation.Easing.Linear,
+	// each nested non-undefined property is converted to a track in a two-frame animation
+	target: { color: { a: 0 } } as MRE.MaterialLike
```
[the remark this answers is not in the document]
stevenvergenz: That type annotation is just for clarity, won't be required. There's a lot of Partial<>s missing in this design.

---
In reply to: [341686022](https://github.com/microsoft/mixed-reality-extension-sdk/pull/464#discussion_r341686022) [](ancestors = 341686022)

design/animation.md
```diff
@@ -0,0 +1,317 @@
...
+const actor = MRE.Actor.CreateFromPrefab(this.context, {
+	prefab: assets.prefabs[0],
+	actor: { name: 'mymodel' }
+});
+```
+
+Loading and playing a native animation from glTF:
+
+```ts
+actor.animations('walk').play();
```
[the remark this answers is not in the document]
stevenvergenz: Oh, apologies! I thought I remembered this already being a field. I'll fix this up.

---
In reply to: [341683554](https://github.com/microsoft/mixed-reality-extension-sdk/pull/464#discussion_r341683554) [](ancestors = 341683554)

design/animation.md
```diff
@@ -0,0 +1,360 @@
...
+in the animation. Instances participate in the patching system, so changing properties on the animation instance
+are pushed down to clients automatically.
+
+An animation instance can be set to start another animation on the instant that it finishes, by setting an instance's
+`continueWith` property. Because it's set in advance, there is zero downtime between one ending and another starting,
+and does not require a round trip.
+
+```ts
+class Animation {
+	bind(...args: Animatible[]): AnimationInstance;
```
eanders-MS: Does this method need to be updated to reflect the name-based binding? I like the name-based binding better than index-based.

design/animation.md
```diff
@@ -0,0 +1,360 @@
...
+		for (const [k,v] of children) {
+			this[k] = v;
+			v.parent = this;
+		}
+	}
+	public toJSON() {
+		// convert tree into string
+	}
+}
+function AnimateActor(name: string) {
```
eanders-MS: The name `AnimateActor` seems overly specific to one use case. Conceivably, this could be used for things other than animations. What about `ReflectActor`.

eanders-MS: I suppose the animation-specific part of the function is that it doesn't expose non-animatable fields.

design/animation.md
```diff
@@ -0,0 +1,360 @@
...
+	/** Playback speed multiplier */
+	speed: number;
+	/** What happens when the animation hits the last frame */
+	wrapMode: WrapMode;
+	/** When multiple animations play together, this is the relative strength of this instance */
+	weight: number;
+	/** The current playback time, based on start time and speed */
+	time: number;
+	/** The list of actors/materials that participate in this instance */
+	targets: Readonly<Animatible[]>;
```
eanders-MS: Is this the map of the instance bindings? Probably needs to be `Readonly<Map<string, Animatable>>`

design/animation.md
```diff
@@ -0,0 +1,360 @@
...
+	1. No easing - Setting fields that cannot be interpolated should also be supported (string, boolean, etc.).
+4. Chaining - Begin an animation immediately after another finishes, or on some other event (behavior tie-in?)
+5. Multiple targets - Animate multiple actors/assets, and multiple properties, in lock-step (equivalent to Unity's
+	animation curves)
+6. Blending - Allow multiple animations to run on the same target simultaneously, and blend the computed destinations
+	into one, proportional to a weight property.
+
+To address this need, I propose the following API:
+
+API Design
```
eanders-MS: Goals: I think we should remove "Chaining" add a more generalized goal: "Control". Support starting and stopping animations together, and trigger animations from other animations.

eanders-MS: From chatting offline: Steven's going to explore the idea of making `AnimationInstance` itself animatable. That way we could start and stop animations from other animations the same way we will be animating actor and materials.

eanders-MS: ... and this would make `continueWith` unnecessary.

design/animation.md
```diff
@@ -0,0 +1,355 @@
...
+			this[k] = v;
+			v.parent = this;
+		}
+	}
+	public toJSON() {
+		// convert tree into string
+	}
+}
+function AnimateActor(name: string) {
+	return new TargetPath<never>(`actor.${name}`, {
```
eanders-MS: I don't think a manually-defined parallel data structure is the right way to implement paths. We should be able to use decorators to build up a parallel data structure programmatically.

design/animation.md
```diff
@@ -0,0 +1,355 @@
...
+When created directly, animations are created as an array of "tracks", each composed of:
+1. A generic path to a field that can be animated (see [Paths](#paths)).
+2. An array of keyframes, each composed of:
+	1. A time value.
+	2. A valid value for the targeted path. Can also be a reference to another actor/material, which will be evaluated
+		at runtime. See [References](#references).
+	3. An optional bezier easing function.
+
+```ts
+type Animatible = Actor | Material | AnimationInstance;
```
eanders-MS: Let's include `SoundInstance` in the `Animatible` set. And let's think through the details of how a keyframe could trigger a sound.

design/animation.md
eanders-MS: I think we should expose non-animatable fields here so that they can be references as destination values in keyframes.

design/animation.md
eanders-MS: I take that back. After some exploratory coding, I think we'll have to go the manual route.

design/animation.md
```diff
@@ -0,0 +1,355 @@
...
+This value is re-evaluated every frame on the client, so matches exactly.
+
+```ts
+assets.createAnimation({
+	targets: {
+		"start": "actor"
+		"end": "actor"
+	},
+	tracks: [{
+		target: AnimateActor("start").transform.local.position,
```
eanders-MS: Before implementing this part, let's discuss the format of the selector string. I want to make sure it is friendly to the addition of expressions later. I'm thinking something like:
* A reference with a binding name of "start": `"actor('bind:start', 'transform.local.position')"`
* A reference to a specific actor: `"actor('id:<guid>', 'transform.local.position')"`

This would be friendly to an expression parser.

eanders-MS: This is nice because it would be parsed as a function with parameters.

PR #466, opened 2019-11-07
Prevent unauthorized users from running certain tests

stevenvergenz:


packages/functional-tests/src/test.ts
```diff
@@ -66,4 +68,15 @@ export abstract class Test {
 				}
 			});
 	}
+
+	public checkPermission(user: User) {
+		if (this.modsOnly) {
```
eanders-MS: Make sure all tests can run in the testbed.

PR #467, opened 2019-11-13
Convert to Typescript 3.7 and ESLint

stevenvergenz:
* Update Typescript dependency to [version 3.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html).
* Replace the deprecated tslint with the new eslint.
    * Port existing tslint rules to eslint (to the extent possible).
* Move some NPM tasks to VSCode tasks (no loss in NPM functionality).


packages/functional-tests/src/tests/stats-test.ts
```diff
@@ -62,6 +62,7 @@ export default class StatsTest extends Test {
 		const plainStats = stats as any;
 		let pp = '';
 		for (const k in plainStats) {
```
eanders-MS: These three lines could be simplified to
```
for (const k of plainStats)
```

packages/functional-tests/src/tests/user-test.ts
```diff
@@ -39,7 +39,7 @@ export default class UserTest extends Test {
 	private formatProperties(props: { [key: string]: string }): string {
 		let output = "";
 		for (const k in props) {
```
eanders-MS: Also could be simplified to
```
for (const v of props)
```

packages/sdk/src/utils/readPath.ts
```diff
@@ -15,7 +15,7 @@ export default function readPath(src: any, dst: any, ...path: string[]) {
 		field = path.shift();
 		validateJsonFieldName(field);
 		if (path.length) {
-			if (!dst.hasOwnProperty(field)) {
+			if (!Object.prototype.hasOwnProperty.call(dst, field)) {
```
eanders-MS: Curious, what was the issue here?

stevenvergenz: `hasOwnProperty` could be an assigned property of the object, instead of being the built-in method.

eanders-MS:
> 		this.leave = this.leave.bind(this);

Curious why this change. This is a common JavaScript code pattern. Was lint flagging it?

---
Refers to: packages/sdk/src/adapters/multipeer/client.ts:61 in a304995. [](commit_id = a30499590cb40cf784de574b9e2890b89923e1e7, deletion_comment = True)

eanders-MS:
> 		this.leave = this.leave.bind(this);

What I was expecting: Convert `leave` to an arrow-method on the class. This should make it unnecessary to call bind.

---
In reply to: [554121429](https://github.com/microsoft/mixed-reality-extension-sdk/pull/467#issuecomment-554121429) [](ancestors = 554121429)

---
Refers to: packages/sdk/src/adapters/multipeer/client.ts:61 in a304995. [](commit_id = a30499590cb40cf784de574b9e2890b89923e1e7, deletion_comment = True)

PR #476, opened 2020-01-28
Convert all our IDs from strings to validated GUIDs

stevenvergenz:


packages/sdk/src/adapters/multipeer/session.ts
```diff
@@ -168,12 +164,13 @@ export class Session extends EventEmitter {
 		} catch { }
 	}
 
-	private setAuthoritativeClient(clientId: string) {
-		if (!this._clientSet[clientId]) {
+	private setAuthoritativeClient(clientId: Guid) {
+		const newAuthority = this._clientSet.get(clientId);
+		if (!newAuthority) {
 			log.error('network', `[ERROR] setAuthoritativeClient: client ${clientId} does not exist.`);
+			throw new Error(`Client ${clientId} does not exist.`);
```
eanders-MS: >throw new Error(`Client ${clientId} does not exist.`); [](start = 3, length = 54)

Ensure this exception is handled and doesn't exit SDK code. Or remove it.

stevenvergenz: Would it be better to simply return? I don't think this case is ever hit, because the app would throw later on in this function if it did.

---
In reply to: [372071873](https://github.com/microsoft/mixed-reality-extension-sdk/pull/476#discussion_r372071873) [](ancestors = 372071873)

eanders-MS: Ya I think just returning would be better. We never hit this that we know of. We don't have a way to recover from this if it were to happen.

---
In reply to: [372075051](https://github.com/microsoft/mixed-reality-extension-sdk/pull/476#discussion_r372075051) [](ancestors = 372075051,372071873)

packages/sdk/src/types/runtime/appearance.ts
```diff
@@ -128,16 +128,16 @@ export class Appearance implements AppearanceLike {
...
 	public set mesh(value) {
-		this.meshId = value && value.id || ZeroGuid;
+		this.meshId = value?.id ?? ZeroGuid;
 	}
 
 	/** @inheritdoc */
 	public get meshId() { return this._meshId; }
 	public set meshId(value) {
-		if (!value || value.startsWith('0000')) {
+		if (!value) {
```
eanders-MS: >value [](start = 7, length = 5)

The logic here used to also check for a non-null value of all zeros. How is that accounted for now?

packages/sdk/src/types/runtime/assets/material.ts
```diff
@@ -67,16 +67,16 @@ export class Material extends Asset implements MaterialLike, Patchable<AssetLike
...
 	public set mainTexture(value) {
-		this.mainTextureId = value && value.id || ZeroGuid;
+		this.mainTextureId = value?.id ?? ZeroGuid;
 	}
 
 	/** @inheritdoc */
 	public get mainTextureId() { return this._mainTextureId; }
 	public set mainTextureId(value) {
-		if (!value || value.startsWith('0000')) {
+		if (!value) {
```
eanders-MS: >f (!value) { [](start = 3, length = 12)

Same question here.

packages/sdk/src/types/runtime/appearance.ts
```diff
@@ -99,16 +99,16 @@ export class Appearance implements AppearanceLike {
...
 	public set material(value) {
-		this.materialId = value && value.id || ZeroGuid;
+		this.materialId = value?.id ?? ZeroGuid;
 	}
 
 	/** @inheritdoc */
 	public get materialId() { return this._materialId; }
 	public set materialId(value) {
-		if (!value || value.startsWith('0000')) {
+		if (!value) {
```
eanders-MS: >if (!value) { [](start = 2, length = 13)

Same question here.

packages/sdk/src/types/runtime/appearance.ts
stevenvergenz: All-zero GUIDs don't need this path, since the "then" part is to use the Zero guid instead.

---
In reply to: [372083083](https://github.com/microsoft/mixed-reality-extension-sdk/pull/476#discussion_r372083083) [](ancestors = 372083083)

PR #478, opened 2020-02-06
Initial design for the pen and interactable behaviors.

tombuMS:


design/interactable.behavior.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Architecture
+
+Create a new behavior that is an `InteractableBehavior` for being able to add to an actor.
+
+``` ts
+/**
+ * Interactable behavior class containing the target behavior actions.
+ */
+export class InteractableBehavior extends TargetBehavior {
```
eanders-MS: How would the app be responsive to state changes (when holding or using changes)? ButtonBehavior has events/callbacks for onButton, onClick, etc. Would it make sense to add callbacks for onHeld, onDropped, etc? Or can this be done from DiscreteAction?

design/pen.behavior.md
```diff
@@ -0,0 +1,60 @@
...
+	private _hold: DiscreteAction = new DiscreteAction();
+	private _drawing: DiscreteAction = new DiscreteAction();
+
+	/** @inheritdoc */
+	public get behaviorType(): BehaviorType { return 'pen'; }
+
+	public get hold() { return this._hold; }
+	public get drawing() { return this._drawing; }
+
+	public onDrawDataReceived((data: Transform[]) => void);
```
eanders-MS: I don't see a material difference between the pen and interactable behaviors. `onDrawDataReceived` would be very useful to have on `InteractableBehavior` (but renamed to something less domain specific). If it were added there, what differences remain?

design/pen.behavior.md
```diff
@@ -0,0 +1,60 @@
...
+The network messages for this behavior will use the standard behavior payloads for actions started and stopped.  The two actions supported 
+by the Pen behavior are hold (begin/end) and drawing (begin/end).
+
+In addition to the standard `PerformAction` payload, there will need to be a second message that is used for sending transform recordings
+up to the MRE app on a set frequeny cadence.  This payload may look something like this:
+
+``` ts
+export type DrawRecording = Payload & {
+	type: 'draw-recording';
+	actorId: drawingActor;
```
eanders-MS: Is `drawingActor` a typo?

design/interactable.behavior.md
eanders-MS: Can I add an `InteractableBehavior` to any actor and it will become grabbable? Same question for `PenBehavior`.

design/pen.behavior.md
sorenhan: I agree - it seems like PenBehavior is InteractableBehavior+wet ink effect+drawDataReceived 
If wet ink effect and drawDataReceived were optional features it could be a single behavior.


---
In reply to: [376066644](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/478#discussion_r376066644) [](ancestors = 376066644)

design/interactable.behavior.md
sorenhan: I'm curious about pros and cons - I can see value in both.

---
In reply to: [376071123](https://github.com/Microsoft/mixed-reality-extension-sdk/pull/478#discussion_r376071123) [](ancestors = 376071123)

design/pen.behavior.md
tombuMS: Soren had liked the idea of a separate more generic behavior.  What I am thinking about more now is having a `ToolBehavior` which the `PenBehavior` extends.  I definitely can see other behaviors wanting to extend something like tool and add additional functionality to the `ToolBehavior`.  I will update the design to reflect that with the `onHeldPathReceived` API in the base class `ToolBehavior`.

design/pen.behavior.md
tombuMS: I needed a better name.  I think it should be the `toolActorId`.

design/interactable.behavior.md
tombuMS: Yes.  The inherent part of the tool behavior is that it is grabbable.  When you add that behavior, we would automatically set the `grabbable` field on the actor to true.  We could add that as a property to the `ToolBehavior` if we need that to be configurable, but not sure how useful the tool would be if it wasn't grabbable.

tombuMS: I didn't show that API, but we provide helper functions for a named API to those discrete actions, but those state changes to come through the actor.  I will add the API for the helper functions like the other behaviors to the design for these.  It was redundant so didn't think it was necessary, but can see how it would be helpful in the design doc.

tombuMS:
> 1. How does this work with the existing Grab? I still don't understand why Grab is not a behavior.

Grab is a core interaction within the MRE system and was decided that it should live outside of the behavior system.  This was to simplify the design of the behavior system for the end user, and to make grabbable be a property of the actor.  This facilitates having any behavior be added to an object and it still be grabbable without having the cumbersome background behavior concept we first had.  

> 2. We should support the streaming of transforms for the grab operation and interactable behavior too.

You already have the transform updates on the actor from the actor update.  You should be able to get this through subscribing to transform updates on the actor while it is grabbed.

> 3. Would it make sense to combine Grab, Interactable, and Pen into a single behavior?

No.  I think that having a general base class now called the `ToolBehavior` is a good design.  We will likely want to add other tools that expand upon the idea of the tool and add additional functionality, such as reload for a gun.  Grab should not be collapsed in to any behavior as the design around grab was that it would not be coupled and restricted to the behavior system.  It simplifies the grab API greatly and how the developer interacts with the behavior system.

> 4. Please add sample code to illustrate usage.

I based this off of other design docs in the repo and didn't see them include examples.  I can add them, but they don't really differ from the usage of other behaviors in the SDK.

Additional points:

I am going to work up and update to the actor API with regards to grab for extracting grab path points.  I will see if that API looks like something that would be beneficial, even though it will just wrap the transform update subscription which can already be done in the SDK now.

eanders-MS:
> You already have the transform updates on the actor from the actor update

Oh, interesting point. I wonder if we don't need onDrawDataReceived and the like at all. High-fidelity updates could be supported by the existing actor subscription / transform update system. I like this.

design/tool.behavior.md
```diff
@@ -0,0 +1,95 @@
...
+	private _holding: DiscreteAction = new DiscreteAction();
+	private _using: DiscreteAction = new DiscreteAction();
+
+	/** @inheritdoc */
+	public get behaviorType(): BehaviorType { return 'tool'; }
+
+	public get holding() { return this._holding; }
+	public get using() { return this._using; }
+
+	/**
```
eanders-MS: I can imagine many cases where it would be useful to get an `Actor`'s transform stream that are completely unrelated to `ToolBehavior`. Because of this, I don't think `onUsingPathReceived` belongs here. I propose adding this API to `Actor` instead:
```ts
public onTransformChanged((path: Transform[]) => void) {
   // Transform changed callback registration
   // Auto-subscribe to transform changes
}
public offTransformChanged((path: Transform[]) => void) {
   // Transform changed callback de-registration
   // Auto-unsubscribe to transform changes if this was the last registered callback
}
```

design/pen.behavior.md
```diff
@@ -0,0 +1,79 @@
...
+path on the client and provide the use path to the MRE app for generating a spline curve asset that could have a mesh to applied to it
+before sending it to all the clients to render. and a mesh to apply to this spline curve.
+
+### Behavior
+``` ts
+/**
+ * Pen behavior class containing the target behavior actions.
+ */
+export class PenBehavior extends ToolBehavior {
+	public drawpointLocalOffset: ScaledTransform;
```
sorenhan: Instead of a transform here, maybe it would be a child actor ID, whose transform we use? (or 0,0,0 if not set). I feel like having to add a local coordinates feels more imprecise. I can't think of other places where we input offsets anywhere in code (other than the actor's own transform, of course), and that's not a pattern I think we should create.

design/tool.behavior.md
```diff
@@ -0,0 +1,95 @@
...
+
+The network messages will use the behavior payloads of SetBehavior and PerformAction payloads the same way all
+behaviors do.  In addition, there is an additional message that would come in the form of the path of the tool
+while it is being used.  This will be sent with a higher fidelity than the normal 10 hz transform updates, and 
+will be in the form of a collection of transform recordings.
+
+The message payload for the using path updates could look like the following:
+
+``` ts
+export type UsingToolPath = Payload & {
```
eanders-MS: Instead of this, I think we should update the actor transform subscription to support high-fidelity.

design/setbehavior.api.md
```diff
@@ -0,0 +1,68 @@
...
+buttonBehavior.onButton('pressed', () => {
+    this.state = 2;
+    this.cycleState();
+});
+buttonBehavior.onHover('exit', () => {
+    this.state = 0;
+    this.cycleState();
+});
+
+this.model.setBehavior(buttonBehavior);
```
eanders-MS: What happens if I add the behavior to more than one actor?

design/pen.behavior.md
tombuMS: We need some kind of offset.  Whether an actor or an offset.  The flexibility of a transform offset is that you do not need to create an additional actor just to be able to use the pen tool.  That seems like unnecessary overhead to me in trying to make an actor behave like a pen.  The local offset is only used if set in this case.  If you are supplying the model, you can make your offset very precise.

tombuMS:
> > You already have the transform updates on the actor from the actor update
> 
> Oh, interesting point. I wonder if we don't need onDrawDataReceived and the like at all. High-fidelity updates could be supported by the existing actor subscription / transform update system. I like this.

I think having this callback is useful though, as it already will provide you paths as chunks rather than you having to write all of that logic yourself in the app by collecting it from the actor update.  I was more trying to point out that we do not need to add this callback for grab to collect grab points because that is a less common thing that I think will be used.  When using the took, the use path is pretty central to the usefulness of a tool.

design/setbehavior.api.md
tombuMS: It would get call backs from each actor when the behavior was triggered.  This was one of the reasons why we initially did this more restrictive API, but it may actually be useful to have this behavior where you could have the `behavior` associated with more then one actor.  I am not sure that this restriction is necessary anymore, hence why I posted this API revision.  I would hope that a developer would understand how instances work in TS such that it would be reasonable expectation that if I gave and instance of something to multiple instances of an other thing, that both things would be acting on that instance.  Right?

design/tool.behavior.md
tombuMS: If you want that to be a change to the update message, I am not against that.  This was added here under the assumption I had that we did not want to overload network traffic with a complete update patch just to implement this behavior.

design/tool.behavior.md
tombuMS: If you want this as an all up addition to actor, I am not opposed to that.  I was recommending it on the tool itself because it logically makes sense to have that API ties to the tool.  I think I would prefer adding this to the actor and still having an API on the pen that wraps this call on the actor.  It is more intuitive and discoverable to the developer to know they are using a pen tool and can get the use path right from it.

eanders-MS: Ok then lets put it on Actor. Some thoughts on implementation:
* I think we can make all transform subscriptions high-fidelity, collecting at 10Hz.
* On the client, it seems like we should collect the transforms in FixedUpdate, so that high-end and low-end clients collect the same number of transforms per update.
* Wondering the best way to package these into a payload. A list of transforms could be added to the ActorPatch.

About this API existing here too .. I think this API should be removed from the behavior for two reasons:
* We should try to avoid defining two APIs that do the same thing.
* These functions are too opinionated. They assume to know when you want the transforms and when you don't. It's better to let the app developer write their own wrappers that encode these kids of assumptions.

design/setbehavior.api.md
eanders-MS: The question is: Is this a good API design? Behaviors are stateful. Is allowing multiple actors affect that state beneficial, innocuous, or harmful? I'm leaning toward harmful.

design/pen.behavior.md
```diff
@@ -0,0 +1,79 @@
...
+There will be a new `PenBehavior` that extends the `ToolBehavior` that will have the ability to generate a wet ink effect along the use 
+path on the client and provide the use path to the MRE app for generating a spline curve asset that could have a mesh to applied to it
+before sending it to all the clients to render. and a mesh to apply to this spline curve.
+
+### Behavior
+``` ts
+/**
+ * Pen behavior class containing the target behavior actions.
+ */
+export class PenBehavior extends ToolBehavior {
```
eanders-MS: You're going to hate me for saying this, but I don't think we need a `PenBehavior`. `ToolBehavior` + transform streaming via Actor gives us everything we need to make a pen behavior. (transforms streaming from a child actor at an offset, if an offset is desired).

design/setbehavior.api.md
tombuMS: We already have at least one other case that Steven is running in to where we want to be able to at the very least pass constructor arguments to the behavior.  I see your point on the potential harmful side effects that this can have, and agree that maybe this isn't the right approach.  Let me think on a better approach to solve this problem and I will put it in to another design PR after the tool design has finalized and I get that work in.

design/tool.behavior.md
tombuMS: Fair point.  I will remove it.

design/pen.behavior.md
tombuMS: I don't hate you, but don't totally agree with you either.  I feel I don't have a compelling enough case though to push it further.  I will remove, but will not forget...

design/tool.behavior.md
tombuMS: Now that I think about it, this is actually a separate API.  This isn't simply a wrapper around the registration of the transform updates.  There is more to it than that.  We may want additional info to come along through the recording.  Also we want latency to be invisible with something like inking and doing the recording in the app will not enable zero latency for the path generation.  We should have an online discussion about this.

design/pen.behavior.md
tombuMS: I actually am circling back to this and going to provide a valid use case for having the pen behavior as well as an API on the pen behavior.  We should discuss real time online to hash this out.

design/tool.behavior.md
eanders-MS: > Also we want latency to be invisible with something like inking and doing the recording in the app will not enable zero latency for the path generation.

Can you clarify your meaning here? We've split the inking operation into two parts: a zero-latency "wet ink" effect, and mesh generation which does have latency. Which part are you referring to here?

Additional info like pen angle, pressure, etc would allow for richer curve rendering. Those could be standard brush inputs. I wonder what kinds of data other pen and paint programs encode into their stroke recordings.

design/pen.behavior.md
sorenhan: I would like to know your use case here, too - I shared Eric's opinion about ToolBehavior+transformStreaming=PenBehavior

design/high.resolution.transform.updates.md
```diff
@@ -0,0 +1,42 @@
...
+    // Disable high resolution updates on the client.
+}
+    
+```
+
+## Network
+
+The actor update structure and payload will now need to be able to accept an array of transforms per update.
+This array would contain one transform in the case of normal resolution transform updates, and more in the
+case that high resolution transform updates were enabled for the actor.
```
stevenvergenz: Why would we need to update the actor with anything but the most recent frame? Seems to me like this could be reduced elsewhere.

design/pen.behavior.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+### Draw Data
+
+``` ts
+export interface DrawDataLike {
+	transform: Transform;
+	// Potentially additional data to come, such as:
+	pressure: number;
+}
```
stevenvergenz: Are we assuming that each DrawData frame is spaced equally apart in time? The speed with which you draw might be important.

design/pen.behavior.md
tombuMS: Had a great call and the usefulness of the pen behavior is a bit more clear.  I have updated the design here for the recording API as discussed in the call, and am further updating based on the results of that meeting.

design/tool.behavior.md
tombuMS: I think after the call, the latency I am concerned about centers around the recording of the transforms at the time that the draw begins on the client, and not after a trip to the server for the perform action call to be made.  I am updating the review based on the chat we had to handle this latency and still support the transform update API on the actor.

design/pen.behavior.md
tombuMS: We have discussed including time stamps in to the draw data.  The requirements there are still in flux.

design/high.resolution.transform.updates.md
tombuMS: We are not updating the actor with anything other than the most recent frame in the 10 hz update cycle.  This is adding in the recording of the transforms over the time between updates messages to allow for higher resolution capture of the movement of an object.  That is why this is an API to register a handler to listen to the last path data.

design/pen.behavior.md
```diff
@@ -29,6 +29,16 @@ export class PenBehavior extends ToolBehavior {
 	}
 ```
 
+### Draw Data
+
+``` ts
+export interface DrawDataLike {
+	transform: Transform;
+	// Potentially additional data to come, such as:
+	pressure: number;
```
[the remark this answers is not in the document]
tombuMS: How so?  I though in our discussion earlier that the pen tool was going to be used for gathering data related to some kind of inking interaction.  Is that not right?

design/high.resolution.transform.updates.md
```diff
@@ -0,0 +1,42 @@
...
+containing the full high resolution path that occured for the actor during the 10 hz regulated network update.
+
+## API
+
+ A new actor API will be exposed to allow a developer to register and de-register for high resolution
+ transform updates.  
+
+``` ts
+/**
+ * Add a callback to the tool behavior to receive path updates for while the tool is being used
```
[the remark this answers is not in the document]
tombuMS: Yes.  The comment just needed to be updated.  Thanks for catching that.

design/pen.behavior.md
```diff
@@ -0,0 +1,99 @@
...
+
+The pen behavior is a method of recording the path by which a user moves a grabbed object that they activate so that various scenearios 
+can be carried out that wish to capture the virtual drawing of a user.  This behavior will carry with it client side visuals
+to ensure low-latency feedback, but will supply the complete curve data needed by the MRE app to acto on once the drawing is done.
+The `PenBehavior` will be an extension of the `ToolBehavior`.
+
+## Architecture
+
+There will be a new `PenBehavior` that extends the `ToolBehavior` that will have the ability to generate a wet ink effect along the use 
+path on the client and provide the use path to the MRE app for generating a spline curve asset that could have a mesh to applied to it
```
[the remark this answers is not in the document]
tombuMS: Sure.  Updated.

design/set.behavior.actor.md
```diff
@@ -0,0 +1,33 @@
...
+## API
+
+### In actor.ts
+``` ts
+/**
+ * Sets the behavior on this actor.
+ * @param behavior The type of behavior to set. Pass null to clear the behavior.
+ * @param initOptions The init options object to pass to the behavior to initialize it.
+ */
+public setBehavior<BehaviorT extends Behavior, InitOptionsT>(
```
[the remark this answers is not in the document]
tombuMS: This was what I had pinged you about earlier to chat about.  This will likely be refined further as I develop this feature.  We should chat offline and I will update the design as this evolves from the tech implementation side of this.  This main point here is that I want the developer to be able to pass in optional config params to the set behavior call.  Still working out how best to accomplish that.

tombuMS:
Closing this design review PR as it has been approved.  Opening a new one from this branch with the implementation in it for review.

PR #482, opened 2020-02-11
Reorganize file structure, consolidate imports

stevenvergenz:
* Reorganize the src folder's contents, breaking up the giant `types` folder by topic. The new folder structure looks like this:
    * `src/actor` - For the actor classes and all their components. Some component types are complicated enough to justify their own subfolders (i.e. `physics`).
    * `src/animation` - For the animation types.
    * `src/asset` - For the asset classes.
    * `src/core` - For system-level classes that still need to be exposed to the user (i.e. `Context`).
    * `src/math` - For all the math types.
    * `src/user` - For the handful of user classes.
    * `src/util` - For miscellaneous types that need to be exported (`WebHost`, `Guid`, `log`).
    * `src/internal` - For all types that are not exported. This includes all the networking types, adapters, and a raft of utility functions. Note: there are some "internal" types not in this folder (e.g. `ActorInternal`), which live in the folder that matches their exposed class. These are still not exposed.
* Consolidate import statements in all files to pull from the `src` and `src/internal` directories instead of from individual files/folders.


eanders-MS:
I noticed the adapters are under "internal". We should publish the adapters. WebHost won't be appropriate for everyone.

eanders-MS:
I like the new layout!

stevenvergenz:
Seems reaosnable. Which classes? Just MultipeerAdapter and WebsocketAdapter?

---
In reply to: [584927450](https://github.com/microsoft/mixed-reality-extension-sdk/pull/482#issuecomment-584927450) [](ancestors = 584927450)

eanders-MS:
Yep those are the only adapters.

---
In reply to: [584936881](https://github.com/microsoft/mixed-reality-extension-sdk/pull/482#issuecomment-584936881) [](ancestors = 584936881,584927450)

PR #483, opened 2020-02-14
Port tests to new animation API

stevenvergenz:


packages/functional-tests/src/tests/clock-sync-test.ts
```diff
@@ -139,16 +139,9 @@ export default class ClockSyncTest extends Test {
...
-							x: (xOffset) * scale,
-							y: (yOffset + i * lineHeight) * scale,
-							z: 0,
-						},
-						scale: { x: scale, y: scale, z: scale }
-					}
-				}
+				x: (xOffset) * scale,
+				y: (yOffset + i * lineHeight) * scale,
+				z: 0,
```
eanders-MS: How does the animation system differentiate this as a position and not a scale?

eanders-MS: Disregard! I see it below.

---
In reply to: [383576683](https://github.com/microsoft/mixed-reality-extension-sdk/pull/483#discussion_r383576683) [](ancestors = 383576683)

PR #484, opened 2020-02-14
Pen and Tool Behaviors

tombuMS:
Adds two new behaviors; the `ToolBehavior` and the `PenBehavior`.  Also adds in new functionality to the DiscreteAction that allows for data to be passed along through the `PerformAction` events, as well as adds an optional `performing` action that happens on a 10 hz update while the action is being performed on the client.

### Still left to do:
- Implement the feature in the client runtime
- Add in functionality to Altspace
- Add in functional test for the `PenBehavior`


design/behavior.actions.md
```diff
@@ -0,0 +1,94 @@
+# Behavior Actions
+
+Behaviors are high level concepts that are made up of one or more actions.  These actions make up the interactions
+that the user makes with the given behavior.  These actions are broken up in to three states: `started` and `stopped`.
```
stevenvergenz: "three"?

design/behavior.actions.md
```diff
@@ -0,0 +1,94 @@
...
+- `stopped` - Fired once when the action is stopped being performed.
+
+**Proposed**: In addition, there is an optional update event that can be registered to the action for getting updates while that
+action is being performed on the client.
+
+## Architecture
+
+### Action State
+``` ts
+export type ActionState
```
stevenvergenz: Are these states or state transitions, a.k.a. events? They seem to be used as both, which is confusing to me.

tombuMS: These are state events.  

- `started` is an event raised when the action is first started
- `performing` is an optional update event that happens while the action is being performed
- `stopped` is an event raised when the action is stopped.

design/behavior.actions.md
tombuMS: Yeah.  I need to update the doc here for this.  Good catch.  The actions are broken up in to three states.

packages/sdk/src/actor/behaviors/action.ts
```diff
@@ -1,86 +1,94 @@
-/*!
```
eanders-MS: Did the whitespace change in this file? Diff is failing.

packages/sdk/src/actor/behaviors/buttonBehavior.ts
```diff
@@ -17,24 +17,20 @@ import { TargetBehavior } from './targetBehavior';
  * Button behavior class containing the target behavior actions.
  */
 export class ButtonBehavior extends TargetBehavior {
-	private _hover: DiscreteAction = new DiscreteAction();
-	private _click: DiscreteAction = new DiscreteAction();
-	private _button: DiscreteAction = new DiscreteAction();
+	private _hover: DiscreteAction<void> = new DiscreteAction();
```
eanders-MS: >DiscreteAction<void> [](start = 17, length = 20)

Given the number of `DiscreteAction<void>`  and `ActionHandler<void>`, it might be worth giving those template parameters a default value of `void`, e.g.: `export class DiscreteAction<ActionDataT = void>`

packages/sdk/src/actor/behaviors/action.ts
```diff
@@ -1,86 +1,94 @@
...
+	'started'?: ActionHandler<ActionDataT>;
+	'stopped'?: ActionHandler<ActionDataT>;
+	'performing'?: ActionHandler<ActionDataT>;
+}
+
+/**
+ * Class that represents a discrete action that can be in one of two states,
+ * started or stopped for each user. @see ActionState
+ */
+export class DiscreteAction<ActionDataT> implements Actionable {
```
eanders-MS: >ActionDataT [](start = 28, length = 11)

Suggest `ActionDataT = void`

packages/sdk/src/actor/behaviors/action.ts
```diff
@@ -1,86 +1,94 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { ActionState, Guid, User } from '../..';
+
+/**
+ * The action handler function type.
+ */
+export type ActionHandler<ActionDataT> = (user: User, actionData?: ActionDataT) => void;
```
eanders-MS: >ActionDataT [](start = 26, length = 11)

Suggest `ActionDataT = void`

packages/sdk/src/actor/behaviors/penBehavior.ts
```diff
@@ -0,0 +1,28 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { BehaviorType, ToolBehavior } from ".";
+import { TransformLike } from "../..";
```
eanders-MS: Use single quotes.

packages/sdk/src/actor/behaviors/buttonBehavior.ts
tombuMS: Didn't even know you could do that in TS.  Learn something new every day.

packages/sdk/src/actor/behaviors/action.ts
tombuMS: No the white space is unchanged.  This happened a couple of times since the file move that Steven had done.

PR #493, opened 2020-03-04
Fix import cycle, lint issues

stevenvergenz:


packages/sdk/src/actor/behaviors/penBehavior.ts
```diff
@@ -2,9 +2,9 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-
-import { BehaviorType, ToolBehavior } from '.';
-import { TransformLike } from '../..';
+import { BehaviorType, TransformLike } from '../..';
+// break import cycle
```
eanders-MS: I'm getting an import cycle here. How are you building the project?

eanders-MS: Oh I meant **not** getting an import cycle here! Comment updated.

PR #502, opened 2020-03-05
Fix compilation on Linux (Docker)

stevenvergenz:
There's a case sensitivity issue.


packages/sdk/src/user/user.ts
```diff
@@ -5,7 +5,7 @@
 
 import { Context, GroupMask, Guid } from '..';
 import { Patchable, readPath, } from '../internal';
-import { UserInternal } from './userinternal';
+import { UserInternal } from './userInternal';
```
eanders-MS: eslint should have a rule for this.

PR #503, opened 2020-03-13
[SDK] Animations part 2

stevenvergenz:
* Created schema in which you can define generic keyframe data.
* Added a `TargetPath` system that type-checks keyframe values against the type of the property being animated.
* Add an `animationData` asset type, with placeholder bindings.
* Add ability to create animations from this generic animation data.
* Remove old animations API, and port tests to new system.


packages/sdk/src/animation/animationData.ts
```diff
@@ -0,0 +1,226 @@
...
+		initialState?: Partial<AnimationLike>
+	): Promise<Animation> {
+		// validate names and types
+		const dataTargets = this.targets();
+		const dataPlaceholders = new Set(Object.keys(dataTargets));
+		const targetIds: { [placeholder: string]: Guid } = {};
+
+		for (const placeholder in targets) {
+			if (!dataPlaceholders.has(placeholder)) {
+				throw new Error(`Animation data "${this.name || this.id}" has no reference ` +
```
eanders-MS: You're introducing a lot of new thrown `Error`s in this changeset. Thrown exceptions aren't a great error propagation mechanism in SDKs because they're not encoded into the function signature in a type-aware way. I think we need a way of returning errors that is reflective (for intellisense), and doesn't require try/catch or null checking everywhere. These new `throw`s can stay in this PR/feature, but let's think of them as stand-ins for a better system. I want to commit to never propagating exceptions from the public API.

I'll do some research and suggest a system we can use across all our public APIs. I'm thinking of something like the `Either` struct from fp-ts: https://gcanti.github.io/fp-ts/modules/Either.ts.html. Essentially, return a value that is either an Error or the result. With some helper methods like `onSuccess`, `onError` added. But let me try some experiments to explore it.

I do think it's appropriate to let platform exceptions propagate. If a file open fails or an exception from another API occurs, let those bubble up to the app. This is really about returning the errors we check for in a uniform, friendly way.

packages/sdk/src/internal/payloads/payloads.ts
```diff
@@ -35,7 +33,7 @@ export type PayloadType
 	| 'animation-update'
 	| 'app2engine-rpc'
 	| 'collision-event-raised'
-	| 'create-animation'
+	| 'create-animation-2'
```
eanders-MS: As I mentioned elsewhere, just reuse `create-animation`. We've so far avoided this sort of versioning.

packages/sdk/src/animation/animation.ts
```diff
@@ -278,4 +338,61 @@ export class Animation implements AnimationLike, Patchable<AnimationLike> {
...
+					});
+				}
+			}
+		})(options.destination, `${typeString}:target`);
+
+		// create the animation data
+		const ac = new AssetContainer(context);
+		const data = ac.createAnimationData('temp', {
+			// force type assumptions
+			tracks: (tracks as unknown) as Array<Track<AnimationProp>>
```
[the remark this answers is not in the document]
stevenvergenz: Because the tracks themselves are not of type `Track`, so I couldn't add them to the array if it were typed.

packages/sdk/src/core/contextInternal.ts
```diff
@@ -205,83 +200,37 @@ export class ContextInternal {
...
-						createdAnim.copy(reply.animations[0]);
-						resolve(createdAnim);
-					} else {
-						reject(reply.result.message);
-					}
-				},
-				reject
-			});
+		const reply = await this.sendPayloadAndGetReply<Payloads.CreateAnimation2, Payloads.ObjectSpawned>({
+			type: 'create-animation-2',
```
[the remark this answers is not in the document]
stevenvergenz: Believe it or not, this feature does not break backward compatibility. Clients on v0.17 support apps all the way down to v0.13. We're removing the APIs from the SDK, but we can't remove them from the plugin yet.

packages/sdk/src/animation/animationData.ts
```diff
@@ -0,0 +1,226 @@
...
+			}, {} as {[placeholder: string]: AnimatibleName});
+	}
+
+	/**
+	 * Bind this animation data to one or more targets to create an [[Animation]].
+	 * @param targets A map of placeholder names to real objects. The names and types must match those in [[targets]].
+	 * @param initialState Initial properties for the new animation.
+	 */
+	public bind(
+		targets: { [placeholder: string]: Animatible },
```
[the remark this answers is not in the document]
stevenvergenz: I'm not opposed to calling it something else, but I think calling it `name` would lead to questions of what exactly is being named. For that I like `placeholder`, because it's clear that it's just a stand-in for something else.

packages/sdk/src/internal/payloads/payloads.ts
stevenvergenz: Let's discuss this in the other convo.

packages/sdk/src/animation/animationData.ts
stevenvergenz: Try-Catch is a cross-language standard, and integrated with the Async-Await flow. I agree that known exception cases should be documented (maybe we can set up a lint rule), but all these other structures seem unnecessary when try-catch accomplishes the same thing in a supported way.

PR #512, opened 2020-03-27
Validate and sanitize network GUIDs

stevenvergenz:


packages/sdk/src/internal/message.ts
```diff
@@ -36,3 +36,31 @@ export type Message<PayloadT = Payloads.Payload> = {
 	 */
 	payload: Partial<PayloadT>;
 };
+
+/**
+ * @hidden
+ * Convert a JSON network message into a Message
+ * @param json
+ */
+export function parseMessage(json: string) {
```
eanders-MS: I don't think this is going to scale. I wonder if we could use JSON Schema here. JSON Schema has a "uri:uuid" format validator we may be able to leverage. We'd define a schema for each payload, validate incoming messages against it, then post process uuid fields to make them lower case.

http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.5

eanders-MS: That said, this is looking like more work than a quick fix. Since we have a workaround (Unreal sending lowercase, hyphenated Guids), let's track this issue in GitHub and we'll prioritize this on the roadmap. Validating incoming messages against a schema is a good idea for many reasons.

packages/sdk/src/internal/payloads/payloads.ts
```diff
@@ -405,3 +407,63 @@ export type DialogResponse = Payload & {
...
+/**
+ * @hidden
+ * Sanitize network input
+ * @param payload A payload fresh from the websocket
+ */
+export function processPayload(payload: Partial<any>) {
+	if (!payload) { return payload; }
+
+	// payload guids
+	if (payload.actorId) { payload.actorId = parseGuid(payload.actorId); }
```
eanders-MS: I'm afraid this isn't a viable solution. This code creates a maintenance problem. We need a solution that doesn't require manual fixup whenever a payload is added or changed. See my JSON Schema suggestion. Also, I don't think we need to tackle this now. We have a workaround.

PR #513, opened 2020-03-27
[design] Adopt glTF's coordinate system

eanders-MS:
Aligning to glTF's coordinate system would be a more consistent design for the SDK, given glTF's prominence in our API. But it could be a lot of work, and there are several approaches we could take. In this design proposal, I want to capture the scope of work and settle on a general approach.


eanders-MS:
this is not happening any time soon. closing for now.

PR #519, opened 2020-04-01
move math utils to a shared package

### LINKED ISSUE #518: Move math and other generic helpers to a separate package
Great work on the `gltf-gen` library. Its perfect for for embedding in other projects and is working brilliantly for me. There is just one frustrating limitation. I can barely believe this is the chain of issues to be honest:

- `gltf-gen` depends on `mixed-reality-extension-sdk`
- `mixed-reality-extension-sdk` depends on `restify`
- `restify` modifies the global `Request` prototype https://github.com/restify/node-restify/issues/1540
- My `express` server (which requires in a module that depends on `gltf-gen`) has the wrong object for `req.query`

This is clearly an issue with `restify` (and has been open there since 2017) but I believe there is an easy fix to be had here. I suggest:

- Create new package in this repo (eg. `shared` or `utils`)
- Move the math classes across to this self-contained package with no dependencies
- `gltf-gen` depends on the shared package (and not the main sdk anymore)

I'm happy to draft up a PR if this is something you would be receptive to. Cheers!

jtfell:
Resolves #518


jtfell:
I'm happy to leave it to you @stevenvergenz - It's your baby and I am really just wanting to be able to import `gltf-gen` without breaking express :)

PR #522, opened 2020-04-03
Move math classes to separate package (stub)

stevenvergenz:


packages/common/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@microsoft/mixed-reality-extension-common",
+  "version": "0.16.1",
```
eanders-MS, line 3: The package version should be something very low, like 0.1.0.

packages/common/package-lock.json
```diff
@@ -0,0 +1,1598 @@
+{
+	"name": "@microsoft/mixed-reality-extension-common",
+	"version": "0.16.1",
```
eanders-MS, line 3: >"version": "0.16.1", [](start = 1, length = 20)

Regenerate package-lock.json to update the version here.

PR #532, opened 2020-04-14
[SDK] Animations overhaul part 2

stevenvergenz:
* Add a whole bunch of tests
* Remove previously deprecated actor functions `create`/`enable`/`disable`/`pause`/`resumeAnimation`, `setAnimationTime`/`State`, and actor events `OnAnimationEnabled`/`Disabled`.
* Deprecate actor functions `createAnimation` and `animateTo` in favor of new versions `AnimationData.bind` and `Animation.AnimateTo`.
* Rename actor properties `animations`/`ByName` to `targetingAnimations`/`ByName`, so it's clear that animations are not owned by actors.
* Add new asset type `AnimationData`, with schema types `TargetPath`, `Track`, and `Keyframe`.
* Remove buggy sync stage for old-style animations.


packages/functional-tests/src/tests/look-at-test.ts
```diff
@@ -7,13 +7,18 @@ import * as MRE from '@microsoft/mixed-reality-extension-sdk';
...
+			{ time: 0, value: { x: 0, y: 0, z: -0.5 } },
+			{ time: 1, value: { x: 1, y: 1, z: -0.5 } },
+			{ time: 2, value: { x: 0, y: 2, z: -0.5 } },
+			{ time: 3, value: { x: -1, y: 1, z: -0.5 } },
+			{ time: 4, value: { x: 0, y: 0, z: -0.5 } }
+		]
+	} as MRE.Track<MRE.Vector3>]
+};
 
 export default class LookAtTest extends Test {
```
eanders-MS, line 23: The monkey head is embedded in the floor for me:
image

packages/functional-tests/src/tests/animation-dynamic-test.ts
```diff
@@ -0,0 +1,93 @@
...
+			time: 2,
+			value: MRE.ActorPath("anchor3").transform.app.position
+		}, {
+			time: 3,
+			value: MRE.ActorPath("anchor1").transform.app.position
+		}]
+	}]
+};
+
+export default class AnimationDynamicTest extends Test {
```
eanders-MS, line 30: The boxes do not move for me:
image

packages/functional-tests/src/tests/animation-scale-test.ts
```diff
@@ -0,0 +1,100 @@
...
+	{ time: 0.67, value: MRE.Quaternion.FromEulerAngles(0, 4 * Math.PI / 3, 0) },
+	{ time: 1, value: MRE.Quaternion.Identity() }];
+const BounceScaleKeyframes: Array<MRE.Keyframe<MRE.Vector3>> = [
+	{ time: 0, value: { x: 1.33, y: 0.667, z: 1.33 }, easing: MRE.AnimationEaseCurves.EaseInQuadratic },
+	{ time: 0.15, value: { x: 0.667, y: 1.33, z: 0.667 } },
+	{ time: 0.85, value: { x: 0.667, y: 1.33, z: 0.667 }, easing: MRE.AnimationEaseCurves.Step },
+	{ time: 1, value: { x: 1.33, y: 0.667, z: 1.33 } },
+];
+
+export default class AnimationNativeTest extends Test {
```
eanders-MS, line 27: I'm seeing around 213 calls to GC.Alloc when this animation plays. That still seems like a lot.

packages/functional-tests/src/tests/animation-dynamic-test.ts
stevenvergenz: The boxes are grabbable, they don't move on their own.

packages/functional-tests/src/tests/animation-scale-test.ts
stevenvergenz: It's a lot, but I believe it's the minimum for this approach. My deep profile says it calls GC.Alloc once per animation, once per animation track, and once each time a JSON field is converted to a float during patching (Json.Net stores values in `object`s, so that's boxing and unboxing penalties). That totals to 7.0KB per frame.
image

If this is too much we can discuss alternative strategies, but the only alternative that comes to mind is much more complicated: a fancy JSON parser that can separate tracks into value types and object types, and interpolate them differently at runtime. If it's even possible, it would take much longer to get right than this implementation.

It's worth noting that this probably is the worst-case scenario, animating 100 numeric fields (position, rotation, and scale on 10 objects). That's probably the total budget for MRE animations in a scene I'd guess.

packages/functional-tests/src/tests/look-at-test.ts
stevenvergenz: This one was simple, the `isPlaying` field wasn't set. There are other tests that are more broken though. Investigating.

packages/functional-tests/src/tests/animation-dynamic-test.ts
eanders-MS: I thought I tried that... will try again.

packages/sdk/src/actor/actor.ts
```diff
@@ -632,83 +631,28 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+	 * @deprecated
+	 * Use [[Animation.AnimateTo]] instead.
+	 * @param value The desired final state of the actor.
 	 * @param duration The length of the interpolation (in seconds).
 	 * @param curve The cubic-bezier curve parameters. @see AnimationEaseCurves for predefined values.
 	 */
 	public animateTo(value: Partial<ActorLike>, duration: number, curve: number[]) {
-		this.context.internal.animateTo(this.id, value, duration, curve);
+		// added this in because it's easy. not so for createAnimation above.
+		Animation.AnimateTo(this.context, this as Actor, {
```
eanders-MS, line 651: Return the promise here?

packages/sdk/src/animation/animationEaseCurves.ts
```diff
@@ -3,39 +3,44 @@
...
 
+/** Control points of a cubic bezier curve. @see [[AnimationEaseCurves]]. */
+export type EaseCurve = [number, number, number, number];
+
 /**
  * Predefined ease curves from https://easings.net/. Curve shapes can be previewed there.
  * Create your own at http://cubic-bezier.com.
  * Note: Curves that end with "Back" include some overshoot.
  */
 export const AnimationEaseCurves = {
```
eanders-MS, line 14: Declaring this as
```ts
export const AnimationEaseCurves: { [index: string]: EaseCurve } = {
```
would make it strongly typed, and make the individual `as EaseCurve` casts unnecessary. In fact it would make it more strongly typed than it is now.

packages/functional-tests/src/tests/index.ts
```diff
@@ -7,7 +7,12 @@ import { TestFactory } from '../test';
 import ActorSpamTest from './actor-spam-test';
 import AltspaceVRLibraryTest from './altspacevr-library-test';
 import AltspaceVRVideoTest from './altspacevr-video-test';
-import AnimationTest from './animation-test';
+import AnimationBlendTest from './animation-blend-test';
```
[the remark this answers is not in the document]
stevenvergenz: `animation-relative` does this. It's kinda boring, but it works.

packages/sdk/src/internal/payloads/payloads.ts
```diff
@@ -35,18 +33,18 @@ export type PayloadType
 	| 'animation-update'
 	| 'app2engine-rpc'
 	| 'collision-event-raised'
-	| 'create-animation'
+	| 'create-animation-2'
```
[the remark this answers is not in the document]
stevenvergenz: Could you clarify what you mean? Runtime versioning is still theoretical AFAIK. Unless v0.17 is going to be a separate runtime from 0.16, we still need this, right? Or were you proposing making 0.17 a breaking change?

packages/sdk/src/animation/animationEaseCurves.ts
stevenvergenz: I agree that from a coding perspective this is preferable. However, if something is declared as a lookup object like that, intellisense doesn't know about its contents, so doesn't autocomplete. Like this it's an inline type, and the keys are known from the type.

eanders-MS: I tested the syntax. Intellisense worked for me. 🤷‍♂️

packages/functional-tests/src/tests/look-at-test.ts
stevenvergenz: Okay, all the tests are fixed. Some of the broken tests actually exposed bugs that I fixed (i.e. setting the transform of an actor with a rigidbody was broken), so that worked out!

packages/sdk/src/animation/animationEaseCurves.ts
stevenvergenz: So how should we resolve this? I spent half my dev cycles without this prop autocompleting until I got around to debugging, so I'm not sure how it works for you.

packages/sdk/src/actor/actor.ts
stevenvergenz: The old method returned `void`, so we should keep that prototype, right?

packages/sdk/src/internal/payloads/payloads.ts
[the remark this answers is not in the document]
eanders-MS: Sure we can keep it for now. Maybe down the road we'll make a breaking change that would warrant cleaning up little things like this.

packages/sdk/src/actor/actor.ts
eanders-MS: I think returning the promise would be an improvement, would you agree? It would allow the caller to await on the completion of the animation.

packages/sdk/src/animation/animationEaseCurves.ts
eanders-MS: It was entirely untyped before. The signature I suggested adds type info. Can you try it locally and see if it works for you?

packages/sdk/src/internal/payloads/payloads.ts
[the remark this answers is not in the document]
stevenvergenz: I think the plan is to remove all these old API artifacts for the 1.0 release.

packages/sdk/src/animation/animationEaseCurves.ts
stevenvergenz: I just tried this again. Added the type declaration `{ [name: string]: EaseCurve }`, recompiled, and Intellisense no longer autocompletes the properties of `MRE.AnimationEaseCurves` in functional tests. Note that even without this explicit type, it is not "untyped", i.e. `any`; it is implicitly typed to `{ Linear: EaseCurve; Step: EaseCurve; ... }`. I suspect that's the info autocomplete is using.

eanders-MS: If my suggestion is isn't working for you then its likely that it won't work for some of our user base either, so let's take your change.

PR #561, opened 2020-04-29
Remove text component from physics planes

eanders-MS:
Recent changes to the text component make it incompatible with having another mesh on the same Actor. This has been the case in Altspace for some time, but a recent change to the MRETestBed.

I also added colors to the physics shapes.

### DIFF
--- a/packages/functional-tests/src/tests/physics-bounce-test.ts
+++ b/packages/functional-tests/src/tests/physics-bounce-test.ts
@@ -7,21 +7,21 @@ import * as MRE from '@microsoft/mixed-reality-extension-sdk';
 
 import { Test } from '../test';
 
-const defaultBallColor = MRE.Color3.FromInts(220, 150, 150);
-
 export default class PhysicsBounceTest extends Test {
 	public expectedResultDescription = "Balls and boxes hit the ground and bounce.";
 	private assets: MRE.AssetContainer;
 	private interval: NodeJS.Timeout;
-	private ballboxMat: MRE.Material;
+	private materials: MRE.Material[] = [];
 	private bouncePlane: MRE.Actor;
 
 	public async run(root: MRE.Actor): Promise<boolean> {
 		this.assets = new MRE.AssetContainer(this.app.context);
 
-		this.ballboxMat = this.assets.createMaterial('ball', {
-			color: defaultBallColor
-		});
+		this.materials.push(this.assets.createMaterial('mat1', { color: MRE.Color3.FromHexString('#ff0000').toColor4() }));
+		this.materials.push(this.assets.createMaterial('mat2', { color: MRE.Color3.FromHexString('#ff7700').toColor4() }));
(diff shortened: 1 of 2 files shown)


packages/functional-tests/src/tests/physics-bounce-test.ts
```diff
@@ -7,21 +7,21 @@ import * as MRE from '@microsoft/mixed-reality-extension-sdk';
...
 import { Test } from '../test';
 
-const defaultBallColor = MRE.Color3.FromInts(220, 150, 150);
-
 export default class PhysicsBounceTest extends Test {
 	public expectedResultDescription = "Balls and boxes hit the ground and bounce.";
 	private assets: MRE.AssetContainer;
 	private interval: NodeJS.Timeout;
-	private ballboxMat: MRE.Material;
+	private materials: MRE.Material[] = [];
```
stevenvergenz, line 14: Isn't this redundant with `this.assets.materials`?

packages/functional-tests/src/tests/physics-friction-test.ts
```diff
@@ -7,21 +7,21 @@ import * as MRE from '@microsoft/mixed-reality-extension-sdk';
...
 import { Test } from '../test';
 
-const defaultBoxColor = MRE.Color3.FromInts(220, 150, 150);
-
 export default class PhysicsFrictionTest extends Test {
 	public expectedResultDescription = "Boxes slide with different frictions.";
 	private assets: MRE.AssetContainer;
 	private interval: NodeJS.Timeout;
-	private boxMat: MRE.Material;
+	private materials: MRE.Material[] = [];
```
stevenvergenz, line 14: Same here.

packages/functional-tests/src/tests/physics-bounce-test.ts
eanders-MS: Not exactly. If there were any more materials defined in the container then this would the subset applicable to the spawned physics objects. I could rename it `palette`.

jabenk:
Thanks!

packages/functional-tests/src/tests/physics-bounce-test.ts
```diff
@@ -7,21 +7,21 @@ import * as MRE from '@microsoft/mixed-reality-extension-sdk';
...
 	private bouncePlane: MRE.Actor;
 
 	public async run(root: MRE.Actor): Promise<boolean> {
 		this.assets = new MRE.AssetContainer(this.app.context);
 
-		this.ballboxMat = this.assets.createMaterial('ball', {
-			color: defaultBallColor
-		});
+		this.materials.push(this.assets.createMaterial('mat1', { color: MRE.Color3.FromHexString('#ff0000').toColor4() }));
```
jabenk, line 20: I am getting here the error " This line has a length of 123. Maximum allowed is 120  max-len". Can this length be adjusted in VS?

PR #563, opened 2020-04-29
Support conditional request headers for static files

stevenvergenz:
Allow cache management through default static file hosting and blob hosting. This includes adding `ETag` headers to responses and returning `304 Not Modified` when appropriate.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1047,6 +1047,23 @@
         "glob-to-regexp": "^0.3.0"
       }
     },
+    "@netflix/nerror": {
+      "version": "1.1.3",
+      "resolved": "https://registry.npmjs.org/@netflix/nerror/-/nerror-1.1.3.tgz",
+      "integrity": "sha512-b+MGNyP9/LXkapreJzNUzcvuzZslj/RGgdVVJ16P2wSlYatfLycPObImqVJSmNAdyeShvNeM/pl3sVZsObFueg==",
+      "requires": {
+        "assert-plus": "^1.0.0",
+        "extsprintf": "^1.4.0",
+        "lodash": "^4.17.15"
+      },
+      "dependencies": {
+        "extsprintf": {
+          "version": "1.4.0",
+          "resolved": "https://registry.npmjs.org/extsprintf/-/extsprintf-1.4.0.tgz",
+          "integrity": "sha1-4mifjzVvrWLMplo6kcXfX5VRaS8="
+        }
+      }
+    },
     "@nodelib/fs.stat": {
       "version": "1.1.3",
       "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-1.1.3.tgz",
@@ -1190,6 +1207,21 @@
       "version": "12.6.8",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-12.6.8.tgz",
       "integrity": "sha512-aX+gFgA5GHcDi89KG5keey2zf0WfZk/HAQotEamsK2kbey+8yGKcson0hbK8E+v0NArlCJQCqMP161YhV6ZXLg==",
+      "dev": true
+    },
+    "@types/restify-errors": {
+      "version": "4.3.3",
+      "resolved": "https://registry.npmjs.org/@types/restify-errors/-/restify-errors-4.3.3.tgz",
+      "integrity": "sha512-kIjSIPHH12R4ivtqnZ0FKpMBt/5EoU2EuGpAF9AcYNGm3cXYDC2foxYlhsN89A1xU5AfnnxhKrHJVXA9Zi/NxQ==",
+      "dev": true,
+      "requires": {
+        "@types/verror": "*"
+      }
+    },
+    "@types/verror": {
+      "version": "1.10.3",
+      "resolved": "https://registry.npmjs.org/@types/verror/-/verror-1.10.3.tgz",
+      "integrity": "sha512-7Jz0MPsW2pWg5dJfEp9nJYI0RDCYfgjg2wIo5HfQ8vOJvUq0/BxT7Mv2wNQvkKBmV9uT++6KF3reMnLmh/0HrA==",
       "dev": true
     },
     "@typescript-eslint/eslint-plugin": {
@@ -1456,8 +1488,7 @@
     "assert-plus": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/assert-plus/-/assert-plus-1.0.0.tgz",
-      "integrity": "sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU=",
-      "dev": true
+      "integrity": "sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU="
     },
     "assign-symbols": {
       "version": "1.0.0",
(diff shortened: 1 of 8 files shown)


packages/sdk/src/util/restify.plugins.serveStaticFiles.d.ts
```diff
@@ -0,0 +1,21 @@
+/*!
```
eanders-MS, line 1: custom .d.ts files should go into a `types` folder outside `src`, and then that folder is added to tsconfig.json's  `types` entry.

packages/sdk/src/util/webHost.ts
```diff
@@ -5,12 +5,20 @@
 
 import { resolve as urlResolve } from 'url';
 import * as Restify from 'restify';
+//import './restify.plugins.serveStaticFiles';
```
eanders-MS, line 8: After moving the .d.ts file to the `types` folder and adding it to tsconfig.json, this import statement should work.

packages/sdk/src/util/webHost.ts
```diff
@@ -67,11 +75,8 @@ export class WebHost {
 			// host static binaries
 			(req, res, next) => this.serveStaticBuffers(req, res, next),
 			// host static files
-			Restify.plugins.serveStatic({
-				directory: this._baseDir,
-				default: 'index.html'
-			}
-			));
+			Restify.plugins.serveStaticFiles(this._baseDir, { index: 'index.html', etag: true }),
+			Restify.plugins.conditionalRequest());
```
eanders-MS, line 79: It isn't clear to me how Restify would be able to check the static buffer's etag, and return a 304/Not Modified if the resource hasn't changed. How does that work?

eanders-MS, line 79: I think we want to use `Cache-Control: no-cache` semantics, meaning that every cache hit on the client would result in an etag check and possible re-download (by sending the `If-None-Match` header with the last known etag). There would be no set `max-age` on the resource. Will the client side need to be updated to send the `If-None-Match` header, and handle a 304/Not Modified result code?

packages/sdk/src/util/webHost.ts
stevenvergenz: It doesn't, because that import statement is looking for a `built` version of that file, and it won't exist. I should remove this.

packages/sdk/src/util/webHost.ts
stevenvergenz: I compute the etag when the buffer is registered, then serve it with requests. That `conditionalRequest` plugin doesn't compute etags, it just looks at the header already set in previous steps.

stevenvergenz: Yes, the client today cannot take advantage of any of this code. I'm working on a design doc for an update client asset cache system, and this will be a dependency of almost any system I cook up.

packages/sdk/src/util/webHost.ts
eanders-MS: If you remove the `./` path portion (and do the other stuff I mentioned), then it should work.

packages/sdk/src/util/webHost.ts
eanders-MS: If the app re-registers an updated buffer, it will get a new etag. Does this invalidate the cached headers in the plugin?

stevenvergenz: Nothing is cached anywhere inside Restify. The only caching is what you see. That said, this implementation was totally broken, and I've rewritten it. TL;DR: Nothing that happens after `res.send` counts, and the conditional request middleware was after the main handler. I split it up.

stevenvergenz: I imagine modifying the client asset cache to store etag and last-modified alongside the processed object. Then when you need the object. attempt the HTTP request first with those headers, then fall back to the cached version on 304.

eanders-MS: Something I'm still not clear on: When a GET request comes in for a static buffer, and the request includes an etag (via If-None-Match header), how exactly is the incoming etag compared with the static buffer's etag in order to know whether to send a 304/Not Modified, or send a 200/OK + the asset?

eanders-MS: Here's a nice overview of control with with etags: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching

stevenvergenz: So there are two routes here, static files and static buffers. Static files have always supported these headers it seems. For static buffers, I had to modify it so the new request flow goes like this:

1. Request for a buffer comes in and hits the `/buffers/:name` route.
2. The first handler in the chain, `checkStaticBuffers`, looks up whether the given buffer exists, and if so, adds the current ETag value to the response, but does not yet *send* the response. This ETag value is computed from the contents of the buffer when it's registered.
3. The second handler, `Restify.plugins.conditionalRequest`, checks the response ETag header against any `If-Match` request headers, and as required sends the `304` response and terminates the handler chain.
4. The third and final handler, `serveStaticBuffers`, is only reached if a full response is required. It sends the remaining response headers and the response body, and closes out the request.

packages/sdk/src/util/webHost.ts
```diff
@@ -63,42 +68,47 @@ export class WebHost {
...
-		// see if there's a handler registered for it
-		if (!procPath || !this.bufferMap[procPath]) {
-			return next();
+	private checkStaticBuffers = (req: Restify.Request, res: Restify.Response, next: Restify.Next) => {
+		const info = this.bufferMap[req.params.name];
+		if (info) {
+			res.setHeader('ETag', info.etag);
+			next();
+		} else {
+			next(new Error(`No buffer registered under name ${req.params.name}`));
```
eanders-MS, line 84: What happens with this error? What result code is returned to the client in this case?

packages/sdk/src/util/webHost.ts
eanders-MS: Thanks for the explanation. This sounds like the right flow.

packages/sdk/types/restify.plugins.serveStaticFiles.d.ts
```diff
@@ -0,0 +1,22 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import * as Restify from 'restify';
+import { Stats } from 'fs';
+
+// TODO: File a pull request adding this to @types/restify
```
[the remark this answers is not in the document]
stevenvergenz: This is addressed in your PR, let's discuss it there.

packages/sdk/src/util/webHost.ts
stevenvergenz: I had thought this would return an http error code, but I'm not positive. Let me test it. Mostly just copied that pattern from a restify sample.

stevenvergenz: It used to throw a 500 error. I've updated it to be a 404.

package-lock.json
```diff
@@ -1047,6 +1047,23 @@
         "glob-to-regexp": "^0.3.0"
       }
     },
+    "@netflix/nerror": {
```
eanders-MS, line 1050: >"@netflix/nerror": [](start = 4, length = 18)

We live in a strange new world.

package.json
```diff
@@ -40,6 +41,7 @@
     "typescript": "^3.7.2"
   },
   "dependencies": {
+    "restify-errors": "^8.0.2",
```
eanders-MS, line 44: This should go in packages/sdk/package.json, not here. Same for @types above.

stevenvergenz: Good call, done.

PR #570, opened 2020-05-01
Initial Tool and Pen Behavior Functionality

tombuMS:
This is the first pass in introducing the 3D inking feature with the Pen Behavior.  This is currently not fully ready and is not exported to be used by developer apps, but the foundational pieces are required for the 2D Surface inking feature that is higher priority at the moment.

All functional tests have been run to ensure no regressions.


packages/sdk/src/actor/actorInternal.ts
```diff
@@ -40,11 +40,11 @@ export class ActorInternal implements InternalPatchable<ActorLike> {
 		const behavior = (this.behavior && this.behavior.behaviorType === actionEvent.behaviorType)
 			? this.behavior : undefined;
 		if (behavior && behavior._supportsAction(actionEvent.actionName)) {
-			behavior._performAction(actionEvent.actionName, actionEvent.actionState, actionEvent.user);
+			behavior._performAction(actionEvent.actionName, actionEvent.actionState, actionEvent.user, actionEvent.actionData);
```
eanders-MS, line 43: This line causes a lint error (max line length). Please run a lint check. There are a few new warnings in this changeset.

packages/sdk/src/actor/behaviors/action.ts
```diff
@@ -1,86 +1,94 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { ActionState, Guid, User } from '../..';
+
+/**
+ * The action handler function type.
+ */
+export type ActionHandler<ActionDataT> = (user: User, actionData?: ActionDataT) => void;
```
eanders-MS, line 11: Make this `<ActionDataT = void>`, and then you can revert all `ActionHandler<void>` back to being `ActionHandler`.

packages/sdk/src/actor/behaviors/behaviorTypes.ts
```diff
@@ -11,4 +11,6 @@ export type BehaviorType
 	= 'none'
 	| 'target'
 	| 'button'
+	| 'tool'
```
eanders-MS, line 14: `BehaviorType` is public, no? Don't add 'tool' or 'pen' values until we're ready to make them available.

packages/sdk/src/actor/behaviors/penBehavior.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/* export */ interface PenEventData {
+    drawData: DrawData[];
+}
+
+/**
+ * Pen behavior class containing the target behavior actions.
+ */
+/* export */ class PenBehavior extends ToolBehavior<PenEventData> {
```
eanders-MS, line 23: I would exclude `PenBehavior` and `ToolBehavior` from this PR. You really just need the 'performing' action on `ButtonBehavior`.

packages/sdk/src/actor/actorInternal.ts
tombuMS: Didn't realize that the lint was separate from build now.

packages/sdk/src/actor/behaviors/penBehavior.ts
tombuMS: The pen and tool code is already working and tested.  I don't see the issue with it being in this PR and not exported.  There are other teams that will want to use this and could simply export the class in their own fork to use it,  It is implemented in the test bed as well.

design/high.resolution.transform.updates.md
```diff
@@ -0,0 +1,42 @@
...
+
+The actor update structure and payload will now need to be able to accept an array of transforms per update.
+This array would contain one transform in the case of normal resolution transform updates, and more in the
+case that high resolution transform updates were enabled for the actor.
+
+## Client Implementation
+
+The client side changes will require a subscription mechanism on the actor runtime instance to signal that 
+high resolution updates need to be gathered for the actor during a fixed update interval and queuded to be 
+sent up with the normal 10 hz update message for the actor.
```
olga-microsoft, line 42: Consider adding a way to request a different update frequency, if the host supports a way to configure this.

design/pen.behavior.md
```diff
@@ -0,0 +1,87 @@
...
+contain the draw data in the optional action data of the `PerformAction` payload.
+
+## Sync layer considerations
+
+Behaviors already have sync layer filtering to ensure that the action messages are for the client they happen on to the MRE app only.  Grab 
+also has special sync filtering based on the client that the grab is happening and the rest of the clients.  
+
+## Client implementation
+
+The client will require the addition of a new pen behavior for the host app to implement, as well as a draw recording system that will facilitate
```
olga-microsoft, line 55: What about client optimizations (pre-processing) of the recorded data?

design/high.resolution.transform.updates.md
tombuMS: Currently the runtime only operates on a 10 hz frequency.  We collect the points based on the fixed update of the host app's update loop, and aggregate in to this 10 hertz update cycle, so the actual fidelity is that of the fixed update.  What would be the desired use case for changing the frequency that the updates come in and/or the frequency by which updates are sent across the network?

design/pen.behavior.md
tombuMS: Could you provide a example(s)?  This is the first pass to get things working, and there will be more design around this feature, so knowing your requirements would certainly be helpful.  😄

packages/sdk/src/actor/behaviors/penBehavior.ts
```diff
@@ -0,0 +1,28 @@
...
+	// private drawOriginActorId: Guid;
+
+	/** @inheritdoc */
+	public get behaviorType(): BehaviorType { return 'pen'; }
```
eanders-MS, line 27: I don't think this will compile. 'pen' is no longer a 'BehaviorType'.

packages/sdk/src/actor/behaviors/penBehavior.ts
eanders-MS: We're going to learn a lot from all the planned changes to ButtonBehavior. Having these behaviors in the codebase might slow you down, or break them. Better to not have to maintain them in parallel if we're not shipping them yet.

eanders-MS, line 23: Declaring the PenBehavior class but not exporting it will generate a lint warning (no-unused-vars).

tombuMS: Making abstract as per our offline conversation.

packages/sdk/src/actor/behaviors/penBehavior.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+interface PenEventData {
+    drawData: DrawData[];
+}
+
+/**
+ * Pen behavior class containing the target behavior actions.
+ */
+export abstract class PenBehavior extends ToolBehavior<PenEventData> {
```
eanders-MS, line 23: Afterthought: Mark ToolBehavior and PenBehavior @hidden so they don't show up in generated documentation (yet).

packages/sdk/src/actor/behaviors/index.ts
```diff
@@ -10,3 +10,5 @@ export * from './behavior';
 export * from './behaviorTypes';
 export * from './buttonBehavior';
 export * from './targetBehavior';
+export * from './toolBehavior';
+export * from './penBehavior';
```
eanders-MS, line 14: @tombuMS I'm getting a lint warning from this line. Can you clean it up when you have a moment?
```
C:\code\Microsoft\mixed-reality-extension-sdk\packages\sdk\src\actor\behaviors\index.ts
  14:31  warning  Newline required at end of file but not found  eol-last
```

packages/sdk/src/actor/behaviors/penBehavior.ts
tombuMS: Good call.  Missed that.  Will add in a new PR.

PR #572, opened 2020-05-05
Cleanup lint warnings

eanders-MS:
* Allow TODO comments for the time being.
* Fix formatting in collider.ts.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -59,7 +59,7 @@ module.exports = {
 		"no-shadow": "warn",
 		"no-unmodified-loop-condition": "error",
 		"no-void": "error",
-		"no-warning-comments": ["warn", {"terms": ["todo", "fixme", "tslint"]}],
+		"no-warning-comments": ["warn", {"terms": ["fixme", "tslint"]}],
 		"prefer-const": "warn",
 		"prefer-regex-literals": "warn",
 		"require-unicode-regexp": "error",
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -59,7 +59,7 @@ module.exports = {
 		"no-shadow": "warn",
 		"no-unmodified-loop-condition": "error",
 		"no-void": "error",
-		"no-warning-comments": ["warn", {"terms": ["todo", "fixme", "tslint"]}],
+		"no-warning-comments": ["warn", {"terms": ["fixme", "tslint"]}],
```
stevenvergenz, line 62: I like the lint warning, to remind us "to do" whatever is mentioned. If there's a TODO in an infrequently reviewed code path, we might not see it or be aware of it.

eanders-MS: We need a clean build. Some reasons:
* As new developers start working in the codebase, if they see a bunch of warnings the first time they build it, they'll never pay attention to the build output again. They won't notice any new warnings their changes might introduce.
* Tracking todos via lint warnings doesn't scale. Actionable todos should be tracked in GitHub issues. That way they're searchable, assignable, taggable, and can be prioritized.

stevenvergenz: I think the intention was that we *shouldn't* let TODOs pile up, but I see your point. Approved.

Issue #576, opened 2020-05-06
[v0.18.n - clock-sync] synchronization failed for user upon re-enter


eanders-MS:
Re-entering the space while the clock-sync test running, the digit strips did not appear. No error message or callstack in the server log.

Issue #529, opened 2020-04-07
Animations using createAnimation fail to sync when user joins late


norybiak:
The new animation system will fail to properly update join-in-progress users.

I found the culprit after some digging: [cacheAnimationCreation() in session.ts](https://github.com/microsoft/mixed-reality-extension-sdk/blob/409bdaf08e7ac0883e20db2ce281670d78245cb9/packages/sdk/src/internal/adapters/multipeer/session.ts#L420) called from ['object-spawned' in rules.ts](https://github.com/microsoft/mixed-reality-extension-sdk/blob/409bdaf08e7ac0883e20db2ce281670d78245cb9/packages/sdk/src/internal/adapters/multipeer/rules.ts#L756).

Looking at the network traffic logs, 'object-spawned' is received some time after the app handles the animation initialization. cacheAnimationCreation() sets anim.update to undefined.

Users that join late try to access this cached update during the [syncing stage](https://github.com/microsoft/mixed-reality-extension-sdk/blob/409bdaf08e7ac0883e20db2ce281670d78245cb9/packages/sdk/src/internal/adapters/multipeer/protocols/clientSync.ts#L233) but the sync payload doesn't get sent since anim.update is undefined.

stevenvergenz:
@norybiak The `update` field is only supposed to contain changes from the default. If it's null, then all clients will get the initialization state, as defined in the `create-animation` message. Still, this does sound real, even if this isn't the culprit. I'll investigate.

stevenvergenz:
The animation functional test syncs for late joiners successfully. Could you share more details of your use case where it fails?

norybiak:
I should have been more specific to which part of the animation system this bug applies to. The syncing issue I'm seeing is when an animation is created using `actor.createAnimation()`. 

I've figured out my problem and it comes down to having to wait for the animation to initialize before trying to update it.

The following is my test that produces the unexpected syncing issue:
```javascript
        myActor.createAnimation('test', 
        {
            keyframes: 
            [{
                time: 0,
                value: { transform: { local: { position: { x: 1 } } } }
            }, {
                time: 10,
                value: { transform: { local: { position: { x: 10 } } } }
            }],
            events: [],
            wrapMode: MRE.AnimationWrapMode.Loop
        });

        myActor.enableAnimation("test");
```

[...]

PR #579, opened 2020-05-07
fix clock-sync-test to await animation binding

norybiak:
First described in https://github.com/microsoft/mixed-reality-extension-sdk/issues/529

I didn't elaborate enough as to why it fails. 

Playing the animation before the animation's ```object-spawned``` payload is returned means that ```animation.update``` is overwritten with undefined.

Simply have to await animData.bind()

https://github.com/microsoft/mixed-reality-extension-sdk/issues/576#issue-613552014


stevenvergenz:
I'd rather fix the sync layer so the API behaves the way we expected. I've got some other anim-related sync stuff to fix, I can take it.

PR #583, opened 2020-05-08
Asset cache design review

stevenvergenz:
Create a new design for caching web-based assets in a safe and reusable manner.

### DIFF
--- a/design/assets.unity.md
+++ b/design/assets.unity.md
@@ -15,7 +15,7 @@ This is a system-wide class instance that is responsible for caching assets beyo
 instance. This could possibly be backed by persistent storage instead of memory. This is primarily intended
 for assets loaded via an HTTP request.
 
-----------------------------
+### StoreAssets
 
 ```cs
 void StoreAssets(
@@ -34,7 +34,7 @@ resource is decremented.
 * `assets` - The collection of assets generated from the given resource
 * `version` - The version of the loaded resource. Will typically be the HTTP response's `ETag` header.
 
------------------------------
+### LeaseAssets
 
 ```cs
 Task<IEnumerable<UnityEngine.Object>> LeaseAssets(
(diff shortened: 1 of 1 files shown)


design/assets.unity.md
```diff
@@ -0,0 +1,157 @@
...
+* `uri` - The resource identifier
+* `assets` - The collection of assets generated from the given resource
+* `version` - The version of the loaded resource. Will typically be the HTTP response's `ETag` header.
+
+### LeaseAssets
+
+```cs
+Task<IEnumerable<UnityEngine.Object>> LeaseAssets(
+	string uri,
+	string ifMatches = null
```
olga-microsoft, line 42: Is this a specific version with a fixed format or just some generic filter? Please update the wording to reflect, what it actually is. If it's a version, then parameter named 'version' would be clearer.

stevenvergenz: You mean update the parameter name? It's already described in the section below.

design/assets.unity.md
```diff
@@ -0,0 +1,157 @@
...
+
+The game object that should be duplicated for new actors.
+
+### GetById
+
+```cs
+AssetMetadata GetById(Guid? id, bool writeSafe = false);
+```
+
+Retrieve an asset by ID. If a write-safe asset is requested, and the stored asset with that ID is shared,
```
olga-microsoft, line 98: Since the behavior is different depending on the parameter, would be better to have 2 methods instead.

eanders-MS: I think a single method makes sense. If an asset is being fetched as writeSafe, it only needs to be duplicated the first time. Subsequent calls will return the already-duplicated asset.

design/assets.unity.md
```diff
@@ -0,0 +1,157 @@
...
+### Unload
+
+```cs
+void Unload(Guid containerId);
+```
+
+Break references to all shared assets and destroy all unshared assets with this container ID.
+
+
+Expected flow: glTF
```
eanders-MS, line 137: I think this flow can be simplified a lot by:
* Disallowing the MRE instance to access the IAssetCache directly, and always go through its AssetManager.
* As I mentioned above, put IAssetCache in charge of fetching the asset from uri.

```C#
// Add an async method to AssetManager that will do the busywork of figuring out whether to look in the global
// cache, whether to copy for writeSafe, etc. Bonus if this can be a template method.
Task<AssetTypeT> AssetManager.GetAsset<AssetTypeT>(Uri uri, string ifVersionMatches = null, bool writeSafe = false);
```
```C#
// Usage example. This will fetch a newer version into the global cache if it was updated remotely.
var gltf = await this.assetManager.GetAsset<AssetTypeT>(uri, ifVersionMatches: null, writeSafe: true);
```

design/assets.unity.md
eanders-MS: How would the MRE runtime know when it is appropriate to provide a value for `ifMatches`?

tombuMS: To Olga's point, wouldn't `matchesVersion` be a more descriptive parameter name as well?

stevenvergenz: @eanders-MS There are two workflows this API supports:
1. Get version from cache, make a `GET` request with the `If-Matches` header, and pull from cache if 304 is returned.
2. Get version from server via `HEAD` request, pull from cache with `ifMatches`, then make a `GET` request if null is returned.

Right now I'm planning on using flow 1, and this argument is for flow 2 as a hedge against future use cases.

@olga-microsoft @tombuMS I named the argument this to make the comparison to the `If-Matches` HTTP header, but if it's confusing I'm happy to rename it.

design/assets.unity.md
```diff
@@ -0,0 +1,157 @@
...
+========================
+
+Goals:
+
+1. Minimize asset downloading and conversion/preprocessing.
+2. Balance memory footprint against load performance, and ensure no unreferenced assets stay in memory.
+2. Maximize asset reuse, so engines can instance and optimize draw calls.
+
+
+IAssetCache
```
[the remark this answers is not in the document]
stevenvergenz: I think loading and caching are sufficiently different that they shouldn't be handled by the same class, though I'm open to discussing a separate proposal to move `AssetLoader.LoadAssets` (not the whole class necessarily) out of the core library.

design/assets.unity.md
stevenvergenz: Accessibility-wise, it's not possible to deny access to the MRE instance, given that the asset manager is *in* the instance, but that's a nitpick. I don't think it's worthwhile to pipe calls to the cache through the asset manager, since they don't share any functionality. The asset manager doesn't have the right information to fetch cached assets either (URL).

design/assets.unity.md
olga-microsoft: Does this parameter follow If-Match specification, can it have multiple values or be asterisk?
I suggest ifMatchesVersion, since you are using 'version' later in the spec to clarify that you are talking about the same string.

stevenvergenz: That seems reasonable.

PR #595, opened 2020-05-20
Add physics bridge

nik0la31:
Use physics bridge to synchronize rigid body transforms
- rigid body transforms are sent as blob to avoid json conversion
- server distributes physics bridge updates between clients without parsing actual transforms


packages/sdk/src/actor/actor.ts
```diff
@@ -72,6 +72,7 @@ export interface ActorLike {
 	 * Any actors parented to this actor will also be exclusive to the given user.
 	 */
 	exclusiveToUser: Guid;
+	owner: Guid;
```
eanders-MS, line 75: When a client that owns an actor disconnects, how is ownership updated? Could it be the switch of authoritative peer picks them up in the client?

packages/sdk/src/internal/adapters/multipeer/rules.ts
```diff
@@ -776,6 +776,53 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
...
+				
+				return message;
+			},
+			beforeReceiveFromClient: (
+				session: Session,
+				client: Client,
+				message: Message<Payloads.ActorUpdate>
+			) => {
+				
+				session.sendPayloadToClients(message.payload, (value) => value.id !== client.id);
```
eanders-MS, line 819: Is this correct: The transforms in this update are limited to actors owned by the sender.

packages/sdk/src/actor/actor.ts
nik0la31: Currently, whole topic around ownership and simulation authority is a big todo.
We'll have to address it more thoroughly. Disconnected client is not covered ATM, and it should next topic to deal.

nik0la31: I am thinking if there is an easy fix to avoid broken state after first commit.

packages/sdk/src/internal/adapters/multipeer/rules.ts
nik0la31: yap, that's correct

PR #596, opened 2020-05-22
Jabenk/mouse pick

jabenk:
- key frame consistently the grabbed objects.


packages/sdk/src/actor/physics/physicsBridge.ts
```diff
@@ -8,6 +8,7 @@ import { TransformLike } from "../transform";
 
 export interface PhysicsBridgeTransformInfo {
```
nik0la31, line 9: I should have deleted whole interface when switched to binary stream for transforms.

jabenk: ah, so this is actually not needed.

PR #607, opened 2020-06-07
Back Compatibility Design

sorenhan:
Finally, here is my draft for the back compatibility design.

### DIFF
--- a/design/back.compatibility.md
+++ b/design/back.compatibility.md
@@ -1,0 +1,79 @@
+## Back Compatibility Plan
+For MRE, back compatibility is an important promise to all stakeholders of the MRE ecosystem. This proposed change will allow MRE client plugins to support multiple "generations" of the MRE SDK. Each generation may have functionality additions or changes in both API surface and protocol, and once a generation goes from preview to finalized release, it's locked for good, and a new generation enters preview phase.
+
+## Stakeholders
+Each stakeholder have a number of different expectations:
+* MRE SDK Core Dev team: I can keep implementing new features or change the protocol at a high pace, without fear of breaking MREs or Host Apps using finalized generations. I can plan a feature set for a new generation, and iterate on it until final release of a generation, without worrying about back compatibility.
+* MRE SDK Users (MRE Devs): Once I deploy my MRE built with a finalized SDK of any generation, it should keep working forever on my target host apps, regardless of host app updates. Even if I rely on undefined behavior, and my app runs by accident, it should keep running forever.
+* Host App Devs: Once I deploy my host app, with the SDK client plugin, it will run all MREs built with a finalized generation of the SDK supported by the client plugin. My app can seamlessly connect to MREs with different SDK generations at the same time - finalized and preview. If I update the client plugin, it should not change behavior on any existing MREs, except may add support for newer generations of MREs. I can update the plugin when I want to, but MRE client plugin updates should never force me to update my app.
+* Non-Unity Client Plugin Devs: Once the Unity client plugin has a finalized release for a generation, I know exactly which features I need to implement to support that generation, and they will never change. I can implement new features and don't have to look back.
+
+In other words, the MRE core dev team will not be able expect other stakeholders to update host apps, alternative client plugins, or deployed MREs ever. And MRE dev team should never even try to fix client plugin bugs that can possibly change functionality in a final version of a generation - MREs may be relying on broken or unidentified behavior, and that's perfectly allowed.
+ 
+
+Note that these rules only apply to the client plugin side, not to the server (SDK) side itself. The server side itself only needs to support a single generation, because MRE devs are free to use any older NPM package from any shipped generation to make their MREs. Therefore it is safe for the core dev team to check in hotfixes for the NPM package; MRE devs are never required to take an update, but if they do they would be expected to test the app.
+
+## Design Options
+
+I considered two basic approaches to implementing multi-generation support
+1. conditional checks sprinkled in whenever changes are applied("if (generation==2) then do A else do B". Pro: it's easy to understand the concept, and easy to start. Con: extremely error prone and hard to test.
+2. complete copy of the generation-specific code (i.e. almost all of the code in the dll). Pro: Minimizes risk of errors. Con: May bloat the plugin size, source tree size, and compile time. Still have to watch out for changes in the wrong version.
+
+The obvious choice here is 2. Approach 1 would causes things to break often for no reason, no matter how careful developers are.
+
+### Source Code Management
+Here are 2 options to managing the source code
+1. move all the generation-specific code into a subfolder per generation - this is where the preview generation is worked on. When finalizing a generation and starting a new preview generation, create a copy of the preview generation folder. Change the namespace in the subfolder. Add a generation-specific prefix on each source file's filename
+2. move all generation-specific code into a separate repository, and have the main repository include this repository as a submodule for each generation.
+
+* Pros for 1: Easy to use with git. 
+* Cons for 1: It's clunky to have a prefix on filenames (but makes it much harder to accidentally modify things in a final generation - many tools highlight filename but not path). It's still possible to modify the wrong generation's files in the repo.
+* Pros for 2: Makes it harder to do accidental checkins in the wrong repo - you commit into a specific version.
+* Cons for 2: git submodule workflow is clunky, not integrated well with visual git tools. While it seems like meging across generations is easier, it's not something we'll see in practice as fixing finalized generations is not something we should be doing.
+
+I recommend version 1. The pain of the git submodule workflow, and loss of productivity is not worth the minor benefits of 2
+
+### Final Binaries
+Here are two options for building the the binaries
+1. Build a single DLL that includes all generations, with a small bit of generation-shared code for connecting to a server, finding the server version, and choosing the generation-specific mre. 
+2. Build one DLLs per generation, plus a management DLL. 
+Pros for 1: it's easier to have a single output DLL than multiple, as build scripts don't need to be updated when a generation is added. Total binary size is smaller than 2, as compiler will probably optimize out many functions that are unchanged across generations. Con: Higher build time for the DLL.
+I recommend option 1. 
+
+
+### Summary of Recommended Design
+We'll duplicate the files into a different subfolder per generation, add a generation-specific prefix to each file, and leave only the absolutely minimal set of files outside this folder. All of the code will be built into a single DLL.
+
+### Example
+For example, imagine the MRE SDK is at version 2.17.0. It would then support 3 generations of MREs: 1.0 (gen 1), 2.0 (gen 2), and 2.17.0 (pre-release for gen 3). 
+Any host app that includes this plugin will get permanent support for gen 1 and 2, and un-guaranteed prototype of gen3. If the host app later updates to SDK version 2.25.0, then gen 1 and gen 2 MREs run unchanged, but MREs built with gen3 version 2.17 to 2.24 may need to be updated 
+
+### Code to be shared across generations
+The more code is duplicated, the less chance of changes having side effects that spill into finalized generations, so the goal is to move as much as possible to be generational. There is *some* code that must be shared, though, in particular anything that interfaces with the main engine.
+
+#### Shared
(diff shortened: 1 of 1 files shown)


design/back.compatibility.md
```diff
@@ -0,0 +1,79 @@
+## Back Compatibility Plan
+For MRE, back compatibility is an important promise to all stakeholders of the MRE ecosystem. This proposed change will allow MRE client plugins to support multiple "generations" of the MRE SDK. Each generation may have functionality additions or changes in both API surface and protocol, and once a generation goes from preview to finalized release, it's locked for good, and a new generation enters preview phase.
```
stevenvergenz, line 2: >once a generation goes from preview to finalized release, it's locked for good, and a new generation enters preview phase [](start = 293, length = 121)

So in the context of this first release, v0 is locked and v1 enters preview?

design/back.compatibility.md
```diff
@@ -0,0 +1,79 @@
...
+### Final Binaries
+Here are two options for building the the binaries
+1. Build a single DLL that includes all generations, with a small bit of generation-shared code for connecting to a server, finding the server version, and choosing the generation-specific mre. 
+2. Build one DLLs per generation, plus a management DLL. 
+Pros for 1: it's easier to have a single output DLL than multiple, as build scripts don't need to be updated when a generation is added. Total binary size is smaller than 2, as compiler will probably optimize out many functions that are unchanged across generations. Con: Higher build time for the DLL.
+I recommend option 1. 
+
+
+### Summary of Recommended Design
+We'll duplicate the files into a different subfolder per generation, add a generation-specific prefix to each file, and leave only the absolutely minimal set of files outside this folder. All of the code will be built into a single DLL.
```
stevenvergenz, line 45: >built into a single DL [](start = 212, length = 22)

How will this work with our dependencies? We update UnityGLTF fairly regularly, and it's not amenable to side-by-side versions. Heck, I'm not sure it's versioned at all.

stevenvergenz, line 45: >built into a single DL [](start = 212, length = 22)

I really hope we can just use "major versions" instead of "generations". Having a .0 version be "finalized" seems very strange from a semver perspective. My proposed alternative: patches only accumulate in the latest major version. So 1.3.17 could be finalized when we cut 2.0.0.

design/back.compatibility.md
```diff
@@ -0,0 +1,79 @@
...
+### Example
+For example, imagine the MRE SDK is at version 2.17.0. It would then support 3 generations of MREs: 1.0 (gen 1), 2.0 (gen 2), and 2.17.0 (pre-release for gen 3). 
+Any host app that includes this plugin will get permanent support for gen 1 and 2, and un-guaranteed prototype of gen3. If the host app later updates to SDK version 2.25.0, then gen 1 and gen 2 MREs run unchanged, but MREs built with gen3 version 2.17 to 2.24 may need to be updated 
+
+### Code to be shared across generations
+The more code is duplicated, the less chance of changes having side effects that spill into finalized generations, so the goal is to move as much as possible to be generational. There is *some* code that must be shared, though, in particular anything that interfaces with the main engine.
+
+#### Shared
+* Initial Connection code. Reason: until host connect to the MRE, it doesn't know the MRE's generation.
+* Code that the host app uses to initialize the MRE system and initialize a single MRE. Reason: host app doesn't need to care about which version each MRE is (MREAppsAPI)
```
stevenvergenz, line 56: >Code that the host app uses to initialize the MRE system and initialize a single MRE [](start = 2, length = 84)

I disagree with this one. We've been making regular updates to the plugin interfaces, so we need to be able to finalize those as well. Maybe we have shared code to determine the version of the MRE at a URL, then duplicate the initialization APIs, and let the host initialize only the generation they want.

design/back.compatibility.md
```diff
@@ -0,0 +1,79 @@
...
+* Host app hooks related to input/user/attachments/createfromlibrary/video player. Reason: the host app generally doesn't care about which version an MRE is, and therefore should call into generation-agnostic hooks. Each generation-agnostic hook should then call into the generation-specific code with the appropriate parameters.
```
stevenvergenz, line 57: >Host app hooks related to input/user/attachments/createfromlibrary/video player. [](start = 2, length = 80)

Same as above.

design/back.compatibility.md
```diff
@@ -0,0 +1,79 @@
...
+Pros for 1: it's easier to have a single output DLL than multiple, as build scripts don't need to be updated when a generation is added. Total binary size is smaller than 2, as compiler will probably optimize out many functions that are unchanged across generations. Con: Higher build time for the DLL.
+I recommend option 1. 
+
+
+### Summary of Recommended Design
+We'll duplicate the files into a different subfolder per generation, add a generation-specific prefix to each file, and leave only the absolutely minimal set of files outside this folder. All of the code will be built into a single DLL.
+
+### Example
+For example, imagine the MRE SDK is at version 2.17.0. It would then support 3 generations of MREs: 1.0 (gen 1), 2.0 (gen 2), and 2.17.0 (pre-release for gen 3). 
+Any host app that includes this plugin will get permanent support for gen 1 and 2, and un-guaranteed prototype of gen3. If the host app later updates to SDK version 2.25.0, then gen 1 and gen 2 MREs run unchanged, but MREs built with gen3 version 2.17 to 2.24 may need to be updated
```
eanders-MS, line 49: In this example, I think that MRE apps running with SDK 2.n.n (2.14, 2.24, etc) should map to the gen2 runtime. If we're developing a prerelease gen3 runtime, the corresponding SDK version should be major version 3, and include a prerelease tag: `3.0.0-prerelease.n`. This frees us up to patch the SDK with the intention of making it more stable against the existing runtime.

design/back.compatibility.md
eanders-MS: I think that SDK 1.1, 1.2, etc should map to the gen1 runtime. Point releases of the SDK should improve stability, not introduce future instability. SDK v2.0.0-prerelease.0 should map to gen2, but may need to be updated when the gen2 runtime is finalized.

design/back.compatibility.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Per generation
+* Anything that touches the scene graph - animation, physics, rendering, geometry, sound, video
+* Anything that touches assets - asset cache, shaders, procedural toolkit (including defaultmaterial)
+* Anything downloading files
+* External libraries that can change over time (including the gltf library)
+* Anything else, really, including utility functions, patching, payloads
+	
+### Steps
+1. Move almost all files into a gen1 folder (everything is considered generational unless explicitly chosen)
```
eanders-MS, line 70: Leveraging Visual Studio Automation may make this process manageable. They provide a pretty good C# API for interfacing with VS projects and files.

design/back.compatibility.md
tombuMS: Totally agree here.  It does not make sense to me that a 2.n.n is mapping to a release ahead.  Having the major release version number map to the generation version is far more understandable and natural in this case.  I don't know any software that does version numbering like this proposal, which already sets this up to be confusing and an outlier.  Usually the additional dot releases for a current version are bug fixes and stability improvements that do not introduce major features or breaking changes.  We should have the ability to patch a release for stability and increment the dot version without anyone running version 2 (gen 2) worrying about any breaking changes or issues.

design/back.compatibility.md
tombuMS: Adding in a third here that the API interface must be versioned in some kind of fashion.

design/back.compatibility.md
tombuMS: I think that some of that list of things will not necessarily need to be versioned from the host app interface perspective.  So I don't agree this is the same as above for all the interfaces here.  I think this section will need to be broken apart to identify which ones need to be versioned and which do not.

design/back.compatibility.md
```diff
@@ -0,0 +1,79 @@
...
+* Cons for 2: git submodule workflow is clunky, not integrated well with visual git tools. While it seems like meging across generations is easier, it's not something we'll see in practice as fixing finalized generations is not something we should be doing.
+
+I recommend version 1. The pain of the git submodule workflow, and loss of productivity is not worth the minor benefits of 2
+
+### Final Binaries
+Here are two options for building the the binaries
+1. Build a single DLL that includes all generations, with a small bit of generation-shared code for connecting to a server, finding the server version, and choosing the generation-specific mre. 
+2. Build one DLLs per generation, plus a management DLL. 
+Pros for 1: it's easier to have a single output DLL than multiple, as build scripts don't need to be updated when a generation is added. Total binary size is smaller than 2, as compiler will probably optimize out many functions that are unchanged across generations. Con: Higher build time for the DLL.
+I recommend option 1.
```
tombuMS, line 41: There is a lot of power and flexibility in having a DLL per major version, as host apps are able to take only the binary versions that they want to support and have full control over the size of their dependencies that they want to include in to their final build.  Not all host apps will want to support all versions going back to V1.  I don't actually think that the assumption that the compiler will do as you think with this in Unity, as this is not a typical use case that IL2CPP would be built specifically to handle.  I think it is better to not make such an assumption in this design, and make it a factor in deciding the method by which we release or final binaries.

This is based off of an offline conversation with @eanders-MS, and I will let him chime in with his thoughts as well.

eanders-MS: It might be a lot simpler to put the put the versioned code into per-version DLLs, with a front-end DLL for the shared parts. One big benefit to this approach is it gives platforms the ability to choose how much backward compatibility to support without still having to take all prior runtime versions.

It would be good to know if il2cpp will do any sort of code coalescing.

design/back.compatibility.md
sorenhan: Yeah, I'll be updating design based on this feedback, so that a final version for generation 1 is 1.n instead of 2.0

design/back.compatibility.md
sorenhan: Changing this slightly, so that 1.n is the final version for generation 1, and 2.0 is first preview of gen 2. That way, it is possible to do hotfixes (1.n+1 etc.) if needed (though as noted elsewhere, hotfixes should be done in a very limited fashion.

design/back.compatibility.md
sorenhan: I don't think I my intent was clear here, and discussing with Tom highlighted that we were in sync. The main motivation here is that host app code should never have to do "if gen 1 then do A, if gen 2 then do B", but that the host app should just interface with the shared code, and the shared code is then responsible for interfacing with the generational code.

I expect the actual code that connects the host app and the shared code to change (and mostly grow) as new features are added for newer generations. In other words, when a host app imports a newer MRE runtime package, code will not just compile until host app code has been fixed up.

design/back.compatibility.md
sorenhan: yeah as I wrote in another comment, I'm happy with that change. Don't know if it has to be 3.0.0-prerelease.n, 3.0.n, or 3.n.m-prerelease, but that's a detail based on what works for NPM, unity plugin packaging etc, I don't have strong feelings either way. I would like to make sure NPM and Unity versions are somewhat obviously linked.

design/back.compatibility.md
sorenhan: Yeah, discussing with Tom, putting a DLL per version, and a shared DLL makes sense, so I'll change this recommendation. At a later point, if total size of all DLLs is a concern, we can investigate if if merging it into a single DLL would strip a lot of unwanted code (I still think it would), but that is an optimization that we don't need to worry about right now.

We'll stick with the plan for having a folder per generation, naming the namespace differently per generation (to make it 100% unambiguous when you see a callstack). We'll create a CSProj per generation DLL, and one for the shared DLL, and then probably have a couple of different solutions - one that includes all generations (for build servers), and one that just includes shared and a specific generation's DLL (speeds up development)

design/back.compatibility.md
sorenhan: Good tip

design/back.compatibility.md
sorenhan: The terms "major version" and "generation" are slightly different to me. Major Version 3 would include generation 1, generation 2, and generation 3 code. I use this term, because "major version" is a generally understood, and I want to make sure any new dev coming in would need to understand the difference.

PR #608, opened 2020-06-08
Add inverted group mask

stevenvergenz:
Add the concept of a GroupMask containing everything *but* the specified groups.

### DIFF
--- a/packages/functional-tests/src/tests/user-mask-test.ts
+++ b/packages/functional-tests/src/tests/user-mask-test.ts
@@ -28,6 +28,11 @@ export default class UserMaskTest extends Test {
 		const red = this.assets.createMaterial('redMat', {
 			color: { r: .854, g: 0.132, b: 0.132 }
 		});
+		const not = this.assets.createMaterial('notMat', {
+			color: MRE.Color3.Gray()
+		});
+		const box = this.assets.createBoxMesh('box', 0.5, 0.5, 0.5);
+		const sphere = this.assets.createSphereMesh('sphere', 0.3);
 
 		// create team labels
 		const textDef = {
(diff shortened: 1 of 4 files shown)


packages/functional-tests/src/tests/user-mask-test.ts
```diff
@@ -73,13 +78,31 @@ export default class UserMaskTest extends Test {
 			}
 		});
 
+		const notRedIcon = MRE.Actor.Create(this.app.context, {
```
eanders-MS, line 81: lint is complaining:
```bash
C:\code\Microsoft\mixed-reality-extension-sdk\packages\functional-tests\src\tests\user-mask-test.ts
   81:9  warning  'notRedIcon' is assigned a value but never used   @typescript-eslint/no-unused-vars
  115:9  warning  'notBlueIcon' is assigned a value but never used  @typescript-eslint/no-unused-vars
```

packages/functional-tests/src/tests/user-mask-test.ts
```diff
@@ -73,13 +78,31 @@ export default class UserMaskTest extends Test {
 			}
 		});
 
+		MRE.Actor.Create(this.app.context, {
+			actor: {
+				name: 'notRedIcon',
+				parentId: root.id,
+				appearance: {
+					enabled: new MRE.InvertedGroupMask(this.app.context, ['red', 'default']),
```
tombuMS, line 86: >InvertedGroupMask [](start = 22, length = 17)

Does this mean not in the default group as well?  Or is every user alway in the default group?

stevenvergenz: Not in the default group, correct.

stevenvergenz: To clarify: every who has no other groups assigned is in the "default" group. So if "default" is in an inverted group mask, then those users will *not* see the given object.

---
In reply to: [437673473](https://github.com/microsoft/mixed-reality-extension-sdk/pull/608#discussion_r437673473) [](ancestors = 437673473)

tombuMS: So then it is possible to use the default group for filtering now as well since you can set and object for the "not default group"?

---
In reply to: [437683618](https://github.com/microsoft/mixed-reality-extension-sdk/pull/608#discussion_r437683618) [](ancestors = 437683618,437673473)

stevenvergenz: You could always use the default group for filtering. Remember, the default group is not all users, it's only users with no other groups assigned. Just now you can apply inverted semantics. It's just like any other group.

---
In reply to: [437688309](https://github.com/microsoft/mixed-reality-extension-sdk/pull/608#discussion_r437688309) [](ancestors = 437688309,437683618,437673473)

PR #620, opened 2020-06-10
user id proposed change in design

sorenhan:
This is a design that will meet my concerns about current user ID design

### DIFF
--- a/design/user-id-update.md
+++ b/design/user-id-update.md
@@ -1,4 +1,4 @@
-## Make User IDs persistent across session ID, but unique per AppID
+## Make User IDs persistent for both registered and unregistered apps, and move make hashing logic into the runtime instead of letting host apps implement it
 
 This is related to github issue #129.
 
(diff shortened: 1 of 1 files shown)


design/user-id-update.md
```diff
@@ -0,0 +1,44 @@
...
+2. Steven wrote this comment : "@eanders-MS This is by design. GDPR requires us to not allow users to be tracked without their permission. As such, any given Altspace user will be assigned a different ID for each instance of a given app. I agree that a persistent ID per app would be tremendously useful, but we have to come up with a strategy to protect our users' privacy first."
+3. Eric wrote about a design:
+	"Registered apps: Same userId for a given user across multiple instances of an app (because the AppId is consistent).
+	Unregistered apps: Different userIds for a given user across multiple instances of an app (because the AppId is derived from the space component id)."
+	"We talked about adding functionality on the website for users to be able to be forgotten by all apps. How this would work: When generating the userId, include a key taken from the user record. When the user chooses to be forgotten, we would roll this key"
+
+Currently in Altspace we use: UserID= combination of altspaceUserID, sessionID, and SCID
+
```
stevenvergenz, line 16: Actually, Altspace only uses the altspaceUserID, and the first non-null string of appID, sessionID, and SCID. AppID here is only defined for registered apps spawned from the world editor.

design/user-id-update.md
```diff
@@ -0,0 +1,44 @@
...
+Currently in testbed we use: UserID=Random generated every connection
+
+### Proposed design
+Just have user runtime hash userID with an AppID, not with sessionID/SCID/anything else. And have the hashing logic be part of the unity runtime instead of be the host app's responsibility
+
+(Note about AppID: AppID for now is URL, but in the future AppID should be either URL or actual AppID registered in the hostapp-agnostic MRE registry - see issue #617)
```
stevenvergenz, line 22: AppID is a long int assigned to the registration by the Altspace backend, not a URL. Unless you're referring to the new registry ID here?

design/user-id-update.md
sorenhan: I updated my design with this note.

design/user-id-update.md
sorenhan: Correct, this was unclear because I was thinking "new registry". Clarified, and made sure we can support both current altspace registry for now and eventual new registry.

design/user-id-update.md
```diff
@@ -0,0 +1,47 @@
...
+	"We talked about adding functionality on the website for users to be able to be forgotten by all apps. How this would work: When generating the userId, include a key taken from the user record. When the user chooses to be forgotten, we would roll this key"
+
+Currently in Altspace we use: UserID= combination of altspaceUserID and either appid (for registered apps) or SCID
+
+Currently in testbed we use: UserID=Random generated every connection
+
+### Proposed design
+1. For all host apps have user runtime hash a persistent userID with an AppID or (if AppID isn't available) URL. 
+2. Have the hashing logic be part of the unity runtime instead of be the host app's responsibility
+3. Provide a workaround that allows multi-instance testing as a single user, by adding an optional postfix to the session ID, which gets hash with the User ID by the client
```
stevenvergenz, line 22: >session ID [](start = 114, length = 10)

As mentioned in Background, the session ID does not contribute to the user ID. Did you mean app ID?

design/user-id-update.md
```diff
@@ -0,0 +1,47 @@
...
+	"Registered apps: Same userId for a given user across multiple instances of an app (because the AppId is consistent).
+	Unregistered apps: Different userIds for a given user across multiple instances of an app (because the AppId is derived from the space component id)."
+	"We talked about adding functionality on the website for users to be able to be forgotten by all apps. How this would work: When generating the userId, include a key taken from the user record. When the user chooses to be forgotten, we would roll this key"
+
+Currently in Altspace we use: UserID= combination of altspaceUserID and either appid (for registered apps) or SCID
+
+Currently in testbed we use: UserID=Random generated every connection
+
+### Proposed design
+1. For all host apps have user runtime hash a persistent userID with an AppID or (if AppID isn't available) URL.
```
stevenvergenz, line 20: >URL [](start = 108, length = 3)

It seems to me that URL is going to be *more* persistent than an app ID, not less. Why bother having an app ID at all at that point? Also, if consistent user IDs are going to be the new norm, we should get a privacy review from Legal.

design/user-id-update.md
```diff
@@ -0,0 +1,47 @@
...
+### Proposed design
+1. For all host apps have user runtime hash a persistent userID with an AppID or (if AppID isn't available) URL. 
+2. Have the hashing logic be part of the unity runtime instead of be the host app's responsibility
+3. Provide a workaround that allows multi-instance testing as a single user, by adding an optional postfix to the session ID, which gets hash with the User ID by the client
+4. By having the hashing logic be in the unity runtime, this is also the natural home for the "me" permission to randomly generate the session ID instead.
+
+(Note about AppID: Currently AppIDs are app specific (Altspace has its' own app registry), but issue #617 is tracking a request for a shared (host app-agnostic) MRE registry. 
+
+### Benefits
+1. users can reliably join same session ID in different spaces. Imagine a high score table or other save game data per user, if altspace space component is destroyed, this data is lost.
```
stevenvergenz, line 28: >session ID [](start = 32, length = 10)

I'm getting really confused at this point. Session IDs should almost always be different per app instance, so you don't get ghost users and other complications. I'm going to assume for the rest of this document you mean app ID?

design/user-id-update.md
sorenhan: correct, this has been cleaned up in updated design

design/user-id-update.md
sorenhan: I don't agree here - once an app is registered, app ID will be permanent, but sometimes it may be necessary to relocate a server and change a URL. That's the reason we went with app IDs to begin with.
About privacy reviews, we will of course make sure we meet all requirements and recommendations

design/user-id-update.md
sorenhan: You are right, this point was not super clear. The point was that you can join the same app from different locations (which would mean different SCID), and your persisted data would still work. Text was updated

design/user-id-update.md
tombuMS: Should we not check with the privacy requirements before making this change?  Seems like a good due diligence step before changing this to establish a persistent user, right?

stevenvergenz: @sorenhan Updating DNS to point to a new server is not hard.

sorenhan: While updating DNS is not hard for many situations, there can easily be cases where that's not sufficient without changing URL - for example moving from Heroku or Openode to Azure, from ngrok to a persistent host, or even when running across multiple separate ngrok sessions. Therefore it *has* to be AppID, to avoid any persisted user-specific data to be lost.

PR #629, opened 2020-06-17
Sleeping bodies message update

jabenk:


jabenk:
Since there is no hurry on this I will first merge this into physics

PR #631, opened 2020-06-22
Adding Actor Attachment Test

patole:
Adding a functional test that tests the actor attachment API

This spawns a large oblong box that attaches to the avatar of the user initiating the test, and two control cubes. The Red/Blue cube cycles between destroying and recreating the attachment when the cube is red, and re-attaching the same actor when the cube is blue. The grey box with the attachment name floating over it cycles between attachment points.

Currently, this passes when the attachment is destroyed every time, but reattaching the same actor leads to improper transforms.

I wasn't able to find anything obvious off the bat in the SDK code that could lead to this, if it makes sense I can open an issue.

### DIFF
--- a/packages/functional-tests/src/tests/actor-attachment-test.ts
+++ b/packages/functional-tests/src/tests/actor-attachment-test.ts
@@ -137,7 +137,7 @@ export default class ActorAttachmentTest extends Test {
 					contents: this.attachments[this.attachmentIndex].toString(),
 					height: 0.1,
 					anchor: MRE.TextAnchorLocation.TopCenter,
-					color: Color3.White()
+					color: MRE.Color3.White()
 				}
 			}
 		});
--- a/packages/functional-tests/src/tests/actor-attachment-test.ts
+++ b/packages/functional-tests/src/tests/actor-attachment-test.ts
@@ -5,35 +5,29 @@
  */
 import * as MRE from '@microsoft/mixed-reality-extension-sdk';
 import { Test } from '../test';
-import { Attachment, AttachPoint, Color3 } from '@microsoft/mixed-reality-extension-sdk';
-import { Box } from '@microsoft/gltf-gen';
-import { create } from 'domain';
-
-
-
 
 export default class ActorAttachmentTest extends Test {
 	public expectedResultDescription = "Actors attaching to avatar points";
 	private assets: MRE.AssetContainer;
 
 	//Limited subset of AttachmentPoints for testing
-	private attachments: AttachPoint[] =[ 
-	 'head'
-	, 'neck'
-	, 'hips'
-	, 'left-eye'
-	, 'left-foot'
-	, 'left-hand'
-	, 'right-eye'
-	, 'right-foot'
-	, 'right-hand'
+	private attachments: MRE.AttachPoint[] =[ 
+		'head',
+		'neck',
+		'hips',
+		'left-eye',
+		'left-foot',
+		'left-hand',
+		'right-eye', 
+		'right-foot', 
+		'right-hand'
 	];
 
 	//Incremented by clicking grey cube
 	private attachmentIndex = 0;
 
 	//Oblong box that gets attached to the avatar of the user running this test
-	private attachedCube : MRE.Actor;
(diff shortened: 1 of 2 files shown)


tombuMS:
One thing to note for working with the SDK codebase, we utilize lint for enforced coding standards.  There are a few warnings and error in the lint for this just so you are aware, because local builds don't run lint unless you use the Build + Lint Tasks in VS Code or run the `npm run build` script from CLI.

packages/functional-tests/src/tests/actor-attachment-test.ts
```diff
@@ -0,0 +1,175 @@
+
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import * as MRE from '@microsoft/mixed-reality-extension-sdk';
+import { Test } from '../test';
+import { Attachment, AttachPoint, Color3 } from '@microsoft/mixed-reality-extension-sdk';
```
stevenvergenz, line 8: >Attachment, AttachPoint, Color3 [](start = 9, length = 31)

These are already imported under the MRE namespace, better not to do it twice.

packages/functional-tests/src/tests/actor-attachment-test.ts
```diff
@@ -0,0 +1,175 @@
...
+import { Box } from '@microsoft/gltf-gen';
+import { create } from 'domain';
```
stevenvergenz, line 10: Are these imports unused?

packages/functional-tests/src/tests/actor-attachment-test.ts
```diff
@@ -0,0 +1,175 @@
...
+				}
+			}
+		});
+		
+		const CreateCube = MRE.Actor.Create(this.app.context, {
+			actor: {
+				name: 'cube1',
+				parentId: this.rootActor.id,
+				appearance: {
+					meshId: this.assets.createBoxMesh('smallBox', 1, 1, 1).id,
```
stevenvergenz, line 109: >this.assets.createBoxMesh('smallBox', 1, 1, 1) [](start = 13, length = 46)

There are a couple duplicates of this "small box" mesh that could be merged into a single reused instance. One pattern I've used before is get-or-create, where you check if there's a mesh by that name in the asset container already. Saves you the trouble of passing the reference around.

packages/functional-tests/src/tests/actor-attachment-test.ts
```diff
@@ -0,0 +1,175 @@
...
+		});
+
+		//Starting state
+		this.createCubeAttachment();
+        this.attachedCube.attach( this.user.id, this.attachments[this.attachmentIndex] );
+
+		await this.stoppedAsync();
+		this.attachedCube.attach(MRE.ZeroGuid, 'none');
+		return true;
+	}
```
stevenvergenz, line 174: This test needs an "unload" method that frees the assets.

stevenvergenz:
My comments are style nitpicks, but since users use these as how they should write their code, it's important. Sorry to nag.

packages/functional-tests/src/tests/actor-attachment-test.ts
patole: Could you provide an example of how that pattern could be integrated into this code? Would using a conditional operator in-line to check for the existence make sense or do we want to provide a different example?

patole: Let me know what you think of the latest iteration

patole:
Comments addressed in latest version, let me know if it all looks good.

packages/functional-tests/src/tests/actor-attachment-test.ts
stevenvergenz: `const boxMesh = this.assets.meshes.find(m => m.name === "smallBox") ?? this.assets.createBoxMesh("smallBox", 1,1,1);` is how I usually do it. This way works too though.

---
In reply to: [445098060](https://github.com/microsoft/mixed-reality-extension-sdk/pull/631#discussion_r445098060) [](ancestors = 445098060)

PR #634, opened 2020-06-24
Add utilities to filter Altspace users by role or event/space

stevenvergenz:
* Add a general-purpose API for writing classes that filter user joins, leaves, and interactions.
* Implement a filter that accepts AltspaceVR moderators.
* Implement a filter that accepts users that are in the same event or space as the first user to join.

### DIFF
--- a/packages/sdk/src/util/userFilter.ts
+++ b/packages/sdk/src/util/userFilter.ts
@@ -13,6 +13,7 @@ export type UserInteractionType = 'joined' | 'input';
 
 /** A class that has user joined/left callback hooks. Applies to [[Context]], and also [[UserFilter]]s. */
 export interface UserEntryExitPoint {
+	users: User[];
 	onUserJoined(callback: UserEntryExitCallback): void;
 	onUserLeft(callback: UserEntryExitCallback): void;
 	offUserJoined(callback: UserEntryExitCallback): void;
@@ -24,14 +25,24 @@ export abstract class UserFilter implements UserEntryExitPoint {
 	private joinedCallbacks = new Set<UserEntryExitCallback>();
 	private leftCallbacks = new Set<UserEntryExitCallback>();
 
-	/** The set of IDs of joined users */
-	protected joinedUsers = new Set<Guid>();
+	/** The set of joined users, indexed by ID */
+	protected joinedUsers = new Map<Guid, User>();
+
+	/** The set of joined users */
+	public get users() {
+		return [...this.joinedUsers.values()];
+	}
 
 	/**
 	 * Set up the user filter
 	 * @param context An MRE.Context object, or another user filter instance
 	 */
 	constructor(protected context: UserEntryExitPoint) {
+		for (const u of context.users) {
+			if (this.shouldForwardUserEvent(u, 'joined')) {
+				this.joinedUsers.set(u.id, u);
+			}
+		}
 		this.context.onUserJoined(u => this.onUpstreamUserJoined(u));
 		this.context.onUserLeft(u => this.onUpstreamUserLeft(u));
 	}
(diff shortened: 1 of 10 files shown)


packages/sdk/src/util/userFilter.ts
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/** Process an input event only from users that pass the filter */
+	public filterInput(eventHandler: ActionHandler<any>): ActionHandler<any> {
+		return (user: User, data: any) => {
+			if (this.shouldForwardUserEvent(user, 'input')) {
+				eventHandler(user, data);
+			}
+		};
+	}
```
tombuMS, line 66: The thing I don't like about this is that it strips away the strongly typed event data this system has been designed to uphold.  Inserting this filter will strip that away from the event handler that the developer is going to supply to this.

tombuMS: Is it not possible to make this a generic function itself and thus propagate the ActionDataT that is originally supplied with the action callback registration method?  Such as `public filterInput<ActionDataT = void>(eventHandler: ActionHandler<ActionDataT>): ActionHandler<ActionDataT>`?

---
In reply to: [445818329](https://github.com/microsoft/mixed-reality-extension-sdk/pull/634#discussion_r445818329) [](ancestors = 445818329)

stevenvergenz: I'll give it a shot.

---
In reply to: [445820208](https://github.com/microsoft/mixed-reality-extension-sdk/pull/634#discussion_r445820208) [](ancestors = 445820208,445818329)

tombuMS: I think that would be `T = void`?  At least that is how we due it in the action handler itself.

---
In reply to: [445832051](https://github.com/microsoft/mixed-reality-extension-sdk/pull/634#discussion_r445832051) [](ancestors = 445832051,445820208,445818329)

tombuMS: Also can you check if the type propagates through to the action handler being supplied to this function?  Just want to see if the type carries through to the developer code.

---
In reply to: [445861996](https://github.com/microsoft/mixed-reality-extension-sdk/pull/634#discussion_r445861996) [](ancestors = 445861996,445832051,445820208,445818329)

stevenvergenz: Good catch with the null vs void, fixed. And I verified that the VS Code Intellisense picked up the type through the filter, though haven't actually run anything.

---
In reply to: [445864069](https://github.com/microsoft/mixed-reality-extension-sdk/pull/634#discussion_r445864069) [](ancestors = 445864069,445861996,445832051,445820208,445818329)

tombuMS: Perfect.  Not too concerned about running.  It should be fine.  More was making sure the intellisense and all was preserved for the event data.  Thanks for updating!

---
In reply to: [445872113](https://github.com/microsoft/mixed-reality-extension-sdk/pull/634#discussion_r445872113) [](ancestors = 445872113,445864069,445861996,445832051,445820208,445818329)

packages/sdk/src/util/userFilter.ts
```diff
@@ -0,0 +1,99 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { ActionHandler, Guid, User } from '..';
+
+/** A callback that handles when a user joins or leaves the MRE session */
+export type UserEntryExitCallback = (user: User) => void;
+
+/** An event type for [[UserFilter.shouldForwardUserEvent]] */
+export type UserInteractionType = 'joined' | 'input';
```
tombuMS, line 12: >input [](start = 46, length = 5)

Is this just here to support the potential need in differentiating if something is an input event in a concrete implementation of the UserFilter predicate method?  I don't see where we ever check against it in the filters implemented.  Just curious if that is the intention.

stevenvergenz: The moderator filter uses this, so moderators can only be added at join and not on click.

---
In reply to: [446271464](https://github.com/microsoft/mixed-reality-extension-sdk/pull/634#discussion_r446271464) [](ancestors = 446271464)

PR #641, opened 2020-06-30
Various server performance optimizations

stevenvergenz:
* Limit frequency of update checks to 50Hz, instead of as fast as the server will run them (~125Hz)
* Limit number of messages that ever go out in a single call to `update`
* Serialize outgoing messages only once, instead of per websocket
* Don't copy messages when they pass through a `Pipe`


packages/sdk/src/core/contextInternal.ts
```diff
@@ -305,7 +305,7 @@ export class ContextInternal {
 
 	public start() {
 		if (!this.interval) {
-			this.interval = setInterval(() => this.update(), 0);
+			this.interval = setInterval(() => this.update(), 20);
```
tombuMS, line 308: >20 [](start = 52, length = 2)

How did we come up with this number?

packages/sdk/src/core/contextInternal.ts
```diff
@@ -345,13 +345,24 @@ export class ContextInternal {
 			...this.animationSet.values()
 		] as Array<Patchable<any>>;
 
+		const maxUpdatesPerTick = 300;
```
tombuMS, line 348: >300 [](start = 28, length = 3)

Is this number high enough for a large scaled app?  We tried something similar to this optimization on Halo 5, and this becomes a case where if your numbers aren't right, you can introduce lag for the objects that didn't make the cut for the last sync.  I just want to make sure we are considering cases where we have a lot of updates coming in for a larger application.

packages/sdk/src/core/contextInternal.ts
stevenvergenz: It's sort of arbitrary. I thought it would be quick enough for most purposes without overwhelming the CPU if checking for updates takes a while.

---
In reply to: [447885855](https://github.com/microsoft/mixed-reality-extension-sdk/pull/641#discussion_r447885855) [](ancestors = 447885855)

packages/sdk/src/core/contextInternal.ts
stevenvergenz: For large-scale apps, it's *especially* important that this number be low. The whole point is to stagger updates over a couple CPU ticks instead of all at once. For MREs, this is a choice between lagging some objects and lagging them all, unfortunately. Because all this is synchronous and networking is async, none of the low-level networking can execute until this method returns.

---
In reply to: [447887863](https://github.com/microsoft/mixed-reality-extension-sdk/pull/641#discussion_r447887863) [](ancestors = 447887863)

packages/sdk/src/core/contextInternal.ts
tombuMS: Have we noticed a problem with the update loop of the server?

---
In reply to: [447921544](https://github.com/microsoft/mixed-reality-extension-sdk/pull/641#discussion_r447921544) [](ancestors = 447921544,447885855)

packages/sdk/src/core/contextInternal.ts
tombuMS: Right, but you are also introducing additional lag by staggering and also by introducing throttling to the update time intervals.  I am not confident that both of these things should be combined in this first optimization.  I am far more in favor of this particular change than to add arbitrary throttling to the update interval of the server.

---
In reply to: [447923762](https://github.com/microsoft/mixed-reality-extension-sdk/pull/641#discussion_r447923762) [](ancestors = 447923762,447887863)

Issue #637, opened 2020-06-25
Physics bridge: Low frequency upload of transforms to server


jabenk:
At the moment the server only routes the transforms from the clients and does not read them (binary format). 
In addition, later the physics bridge might stream via a different server or directly peer to peer.

So we need to create a low-frequency upload channel from each client to the server to update the transforms.

jabenk:
https://github.com/microsoft/mixed-reality-extension-unity/pull/204
https://github.com/microsoft/mixed-reality-extension-sdk/pull/642

PR #642, opened 2020-06-30
Jabenk/mouse pick

jabenk:
Adding a low-frequency server transform upload stream. The physics bridge updates are not readable for the server.
For late joining clients it is important to have the current positions on the server, this should not happen based on subscriptions.
microsoft/mixed-reality-extension-sdk#637


packages/sdk/src/internal/connection/webSocket.ts
```diff
@@ -35,13 +35,13 @@ export class WebSocket extends EventedConnection {
...
 				// NOTE: This will sometimes change message ordering.
-				// setTimeout(() => {
+				//setTimeout(() => {
 				try {
 					super.recv(message);
 				} catch (e) {
 					log.error('network', e);
 				}
-				// }, 250 * Math.random());
+				//}, 200 + 50 * Math.random());
```
nik0la31, line 44: revert file

Issue #610, opened 2020-06-09
Optimize bandwidth


nik0la31:
* don't send updates for sleeping rigid bodies
* don't send snapshots from clients with 0 rigid bodies
* use int for rigid body IDs (instead guid)
* try 30 updates per second (instead 60)

jabenk:
The first 2 items are implemented in the physics branch

PR #648, opened 2020-07-01
Jabenk/devwork

jabenk:
Add a separate subscription for the velocities of the rigid body.


jabenk:
https://github.com/microsoft/mixed-reality-extension-sdk/issues/610

PR #672, opened 2020-07-23
Adding FontTest to test Altspace's fonts with a variety of Unicode ranges

patole:
Test is as simple as start and observe. All our fonts are stress-tested with Latin, Korean, Japanese, and Chinese characters, which cycle automatically every 3 seconds.

### DIFF
--- a/packages/functional-tests/src/tests/font-test.ts
+++ b/packages/functional-tests/src/tests/font-test.ts
@@ -6,14 +6,13 @@
 import * as MRE from '@microsoft/mixed-reality-extension-sdk';
 
 import { Test } from '../test';
-import { Vector3 } from '@microsoft/mixed-reality-extension-sdk';
-import { stringify } from 'querystring';
-import { int } from '../../../sdk/node_modules/@microsoft/mixed-reality-extension-common/built/math/types';
+import { Vector3, Color3 } from '@microsoft/mixed-reality-extension-sdk';
 
 const options = {
 	font: Object.keys(MRE.TextFontFamily) as MRE.TextFontFamily[],
 };
 
+// };
 /**
  * Test the text api functionality
  */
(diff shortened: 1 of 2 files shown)


packages/functional-tests/src/tests/font-test.ts
```diff
@@ -0,0 +1,221 @@
...
+			{blockStart: 0x20, blockEnd:0x07E, name:"Latin Basic"},
+			{blockStart: 0xA0, blockEnd: 0xFF, name:"Latin Suppliment"},
+			{blockStart: 0x100, blockEnd:0x017F, name:"Latin Extended - A"},
+			{blockStart: 0x3060, blockEnd:0x309F, name:"Japanese Hiragana/Katakana"},
+			{blockStart: 0x4E00, blockEnd:0x4E7F, name:"Chinese/Japanese/Korean Ideographs"});		
+
+		this.textBlocks.forEach((v,i,a) =>{ this.createDisplayString(v); });
+
+
+		const position = new Vector3(0,2.5,-.3);
```
tombuMS, line 78: >Vector3 [](start = 23, length = 7)

Prefix with MRE and remove the extra include to keep it consistent with other tests

packages/functional-tests/src/tests/font-test.ts
```diff
@@ -0,0 +1,221 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import * as MRE from '@microsoft/mixed-reality-extension-sdk';
+
+import { Test } from '../test';
+import { Vector3 } from '@microsoft/mixed-reality-extension-sdk';
```
tombuMS, line 9: Remove in favor of the MRE one above.

packages/functional-tests/src/tests/font-test.ts
patole: Yeah I need to find the setting for these auto imports, every time I try the lazy way without  the MRE. it adds the import and I forget to double check.

PR #702, opened 2020-08-31
Initial app-to-app design proposal

in2dair:
This is the initial design review for a photo sharing feature in MRE.


design/app-to-app.md
```diff
@@ -0,0 +1,80 @@
...
+    // [... Deregister listener ...]
+    return true; // If successful
+}
+```
+
+Implementation Suggestions
+------------------
+1. Modify `context` to expose callback registration mechanism
+2. Adjust `contextInternal` and perhaps `multiPeerAdapter` to message the desire for each `<MessageType,SessionId,UserId>` callback to each client (and corresponding remove messages). For callbacks registered for all sessions or all users, this my be sent on connect
+3. Implement the "Photo" message type. Send and receive encoded photos via the existing websocket used by multiPeerAdapter. Future message types (e.g., streaming video) may need to establish related side-channels.
```
in2dair, line 80: Open to tweaking this if folks disagree. I think the occasional photo will be rare and lightweight, but we'd need a case-by-case for future types (and i think it's ok if they use different channels for future streaming types)

design/app-to-app.md
```diff
@@ -0,0 +1,80 @@
+App to App Interactions
+===========
+
+This document describes an extension to the SDK to allow MRE apps to receive content (e.g., photos) from the MRE clients. Content may be generated by the client itself, or, in future iterations not yet described here, may arrive from other apps with coordination from a particular client.
+
+API Design
+-----------
+This design introduces the concept of a `SharingMessage`, a base class of content-specific messages, such as `PhotoSharingMessage`. We provide an outline for the implementation of `PhotoSharingMessage` only, but expect to add more sharing capabilities over time.
+
+`SharingMessage` is itself relatively straightforward, but contains enough information to allow the recipient (callback listener) to cast it to the appropriate type (e.g., `PhotoSharingMessage`).
```
in2dair, line 10: Maybe add something here about why: we envision multiple message types (and perhaps related types) and don't want to add a registration api for each

shanewhitfield: I agree. Ensuring coverage of who, what, why provides a lot of clarity.

design/app-to-app.md
```diff
@@ -0,0 +1,80 @@
...
+ * Returns true if the callback was successfully registered.
+ * @event
+ */
+public onSharingMessage(sessionId: string, userId: Guid, callbacks: SharingCallbacks): bool {
+    // [... Register listener ...]
+    return true; // If successful
+}
+
+/**
+ * Removes the onSharingMessage callback, if any, associated with the given sessio
```
in2dair, line 64: typo: session

design/app-to-app.md
```diff
@@ -0,0 +1,80 @@
...
+class PhotoSharingMessage extends SharingMessage {
+    photoType: Readonly<string>; // Supported values: "jpg" and "png"
+    data: Readonly<Buffer>; // Raw bytes of an image file in the photoType format above
+}
+```
+
+`SharingCallbacks` are a relatively straightforward callback registration mechanism, with an optional mechanism to allow the callback registrant to stop processing of messages and construction of payloads in cases where a particular message would not be acted on.
+
+```ts
+interface IsMessageDesiredCallback {
```
in2dair, line 28: // Fires before the MessageReceivedCallback. When the implementation returns false, may short-circuit additional
// processing on the message. Note that this callback is optional.

design/app-to-app.md
```diff
@@ -0,0 +1,80 @@
...
+interface IsMessageDesiredCallback { 
+    (sessionId: string, userId: Guid) : bool
+}
+
+interface MessageReceivedCallback {
+    (sharingMessage: SharingMessage) : void
+}
+
+class SharingCallbacks {
+    public isMessageDesired : IsMessageDesiredCallback;
```
in2dair, line 37: // Fires before the MessageReceivedCallback. When the implementation returns false, may short-circuit additional
// processing on the message. Note that this callback is optional.

design/app-to-app.md
```diff
@@ -0,0 +1,80 @@
...
+    public messageReceived : MessageReceivedCallback;
+}
+```
+
+Callbacks are registered (and deregistered) via new functions on the context described below. Optionally, callbacks may be scoped to a particular session id or user id. When no session or user id is provided, sharing data will be accepted from any session and user.
+
+For the sake of simplicity, only one callback may be registered for a given ```<MessageType,SessionId,UserId>``` tuple at a time.
```
in2dair, line 44: @shanewhitfield  suggested we expand the reasons for this simplifying restriction. I'll add more to the doc.

shanewhitfield, line 44: Can you add an explanation of why it's ok to accept the simplification's trade off, "only one call back can be registered".

design/app-to-app.md
```diff
@@ -0,0 +1,80 @@
...
+
+```ts
+class SharingMessage {
+    messageType: Readonly<string>;
+    sessionId: Readonly<string>;
+    userId: Readonly<Guid>;
+}
+
+class PhotoSharingMessage extends SharingMessage {
+    photoType: Readonly<string>; // Supported values: "jpg" and "png"
```
kaelspencer, line 20: Is a string the right way to go here? How will the difference between "jpg" and "jpeg" be handled?

in2dair: Good question: As written, "jpeg" wouldn't be accepted. Open to suggestions. Could add it to the list, note explicitly that the client implementation should use "jpg" rather than "jpeg", or keep it as is.

design/app-to-app.md
```diff
@@ -0,0 +1,87 @@
+App to App Interactions
+===========
+
+This document describes an extension to the SDK to allow MRE apps to receive content (e.g., photos) from the MRE clients. Content may be generated by the client itself, or, in future iterations not yet described here, may arrive from other apps with coordination from a particular client.
```
stevenvergenz, line 4: Please describe why this feature is useful, expected, or otherwise adds to the MRE SDK's value.

design/app-to-app.md
```diff
@@ -0,0 +1,87 @@
...
+
+## API Design
+This design introduces the concept of a `SharingMessage`, a base class of content-specific messages, such as `PhotoSharingMessage`. We provide an outline for the implementation of `PhotoSharingMessage` only, but expect to add more sharing capabilities over time.
+
+`SharingMessage` is itself relatively straightforward, but contains enough information to allow the recipient (callback listener) to cast it to the appropriate type (e.g., `PhotoSharingMessage`). We envision multiple message types and don't want to add a registration api for each.
+
+```ts
+class SharingMessage {
+    messageType: Readonly<string>;
+    sessionId: Readonly<string>;
```
stevenvergenz, line 14: Session ID is redundant here, because it is already associated with the websocket that the message arrives on.

design/app-to-app.md
```diff
@@ -0,0 +1,87 @@
...
+    userId: Readonly<Guid>;
+}
+
+class PhotoSharingMessage extends SharingMessage {
```
thespivey, line 18: Should we support sharing a collection of multiple photos in a single message?

design/app-to-app.md
```diff
@@ -0,0 +1,87 @@
...
+    photoType: Readonly<string>; // Supported values: "jpg" and "png"
+    data: Readonly<Buffer>; // Raw bytes of an image file in the photoType format above
```
thespivey, line 20: If the photo can be large, I wonder if we would want to provide a thumbnail here with an async "get me the full file" function.  Maybe even a "get me the file at resolution X in format Y" function?  Goal being to enable the MRE to rapidly respond to photo sharing with some UI and to show progress indicators while the full transfer happens.

stevenvergenz, line 20: It could cause significant server load to deserialize tens of megabytes of JSON, even if most of it is binary. I'm not sure what the right solution is here. Send up a URL that the server would download the image from? Send up a "switch to websocket binary protocol" type message, and stream in the raw bytes there?

design/app-to-app.md
```diff
@@ -0,0 +1,87 @@
...
+ * Registers a set of callbacks invoked at various points in the processing
+ * of a sharing message. Only one callback may be registered at a time for a 
+ * given context and user.
+ * 
+ * Use ZeroGuid to register for all users.
+ * 
+ * Returns true if the callback was successfully registered.
+ * @event
+ */
+public onSharingMessage(userId: Guid, callbacks: SharingCallbacks): bool {
```
stevenvergenz, line 66: Would you want to listen for uploads only from certain users? I would expect the listen function to take just a callback, and the callback receiving the user ID.

design/app-to-app.md
```diff
@@ -0,0 +1,87 @@
...
+public offSharingMessage(userId: Guid): bool {
+    // [... Deregister listener ...]
+    return true; // If successful
+}
+```
+
+## Implementation Suggestions
+
+1. Modify `context` to expose callback registration mechanism
+2. Adjust `contextInternal` and perhaps `multiPeerAdapter` to message the desire for each MessageType callback to each client (and corresponding remove messages). For callbacks registered for all sessions or all users, this my be sent after connect
```
stevenvergenz, line 86: >Adjust `contextInternal` and perhaps `multiPeerAdapter` to message the desire for each MessageType callback to each client (and corresponding remove messages). [](start = 3, length = 159)

I think it would make the most sense for an app to declare its acceptance of shares in the manifest, which is downloaded before the MRE even connects. Or would an app send individual users "ready to receive" type messages?

stevenvergenz:
Please add an expected client/server message flow to this document. It's unclear to me how a transfer would be initiated from what's here.

stevenvergenz (changes requested):
:clock1:

design/app-to-app.md
tombuMS: Concern here is around the above mentioned deserialization.  This could cause a heavy load on the server that would affect other messages coming across the websocket.  We will need to be mindful of the best way to avoid this

design/app-to-app.md
```diff
@@ -0,0 +1,87 @@
...
+For the sake of simplicity, only one callback may be registered for a given ```<MessageType,UserId>``` on a particular session context at a time. This restriction may allow for a more simple bookeeping implementation for registered callbacks within the the sharing implementation on both the app side and the client runtime See [Implementation Suggestions](#Implementation-Suggestions)
+
+```ts
+// Modifications to
+// https://github.com/microsoft/mixed-reality-extension-sdk/blob/master/packages/sdk/src/core/context.ts
+
+/**
+ * Registers a set of callbacks invoked at various points in the processing
+ * of a sharing message. Only one callback may be registered at a time for a 
+ * given context and user.
```
tombuMS, line 59: What's the background on this restriction.

design/app-to-app.md
```diff
@@ -0,0 +1,87 @@
...
+ * 
+ * Returns true if the callback was successfully unregistered.
+ * @event
+ */
+public offSharingMessage(userId: Guid): bool {
+    // [... Deregister listener ...]
+    return true; // If successful
+}
+```
+
```
tombuMS, line 82: - [ ] Needs a client implementation section to detail out proposed changes to client design to support this feature.  

- [ ] Also add in a networking section to outline the payloads being added, as well as how the network communication flow will take place.

tombuMS (changes requested):

design/app-to-app.md
in2dair: The reason for including it here is that a callback instance might be registered across sessions.

design/app-to-app.md
in2dair: I'm intrigued by the switch to binary idea since the url approach might have connectivity and reliability issues... i'll look in to that.

design/app-to-app.md
in2dair: described on line 50 of the current commit (the text above the code block). Note that this restriction is forward compatible with a change in position here; i.e., we could allow multiple registration in the future if we change our minds.

design/app-to-app.md
in2dair: good call. I'll explain the photo scenario.

design/app-to-app.md
in2dair: That's a fair point. The callback itself does contain the user id and there's a mechanism to short circuit the process if the app doesn't want the data for that user...

design/app-to-app.md
in2dair: I'll add plumbing for the size of the image, and allow the app to act on that in the `IsMessageDesired` stage. For now, skipping any app-direct resizing of the image.

in2dair: Let me try to address this by tweaking the IsMessageDesired implementation to expose a lot more information about the payload to the app, like the size, and allow the app to reject the payload before the payload is sent. In addition, I'll tweak the actual sending to specify that we use the binary protocol.

Down the road, I think we can tweak extend this initial message further to contain enough information to bootstrap, say, a video sharing side channel.

design/app-to-app.md
in2dair: I think we need the "is message desired" (what you call "ready to receive") per message thing. I'll flush this out, there's other feedback in this pr on that topic for sure.
I'm not sure we need manifest entries but happy to add it if there's a compelling reason happy to add them.

design/app-to-app.md
in2dair: TODO : change to MIME Type. And note that any other types (e.g., heic) would be on the mre host to convert.

design/app-to-app.md
thespivey, line 66: onSharingMessage does not take a type, it looks like it registers for all SharingMessages.  In a future where there are multiple types of sharing (photo, video, GLTF, text, etc), how will an MRE express the types that it supports?

I see 3 obvious techniques:

  1.  Realtime filtering:  the MRE could receive the message type in IsMessageDesired and reject unsupported sharing types. 

   2.  Runtime type registration:  onSharingMessage takes a specific type or maybe a list of supported sharing message types.

   3.  Manifested type registration:  the manifest.json lists the supported types.

I am just not sure that #1 is ideal.  If the host wants to generate a list of viable targets for a given sharing action, it would need to broadcast this message to every running MRE.  But sending a sharing message to an MRE the user did not explicitly select feels like a privacy risk, we'd have to be very careful with the metadata.  This is also unlikely to be compatible with OS-provided sharing constructs, which expect registration ahead of time.

#2 invites seems well aligned with supporting future scenarios.  For instance, messaging apps commonly register specific sharing targets/shortcuts for recent conversations in addition to their global sharing target.  It seems like a runtime registration design would be needed to support dynamic/multiple registrations of that sort in the future.

design/app-to-app.md
in2dair: Good point. I'll add a transmissionMethod to the data flow outline section i'm adding so that we have a more future proof way of switching to an alternative websocket for example.

design/app-to-app.md
in2dair: Removed user specific registration. added message-type specific registration (e.g., "image") so that an app can use different callbacks for different message types.

design/app-to-app.md
in2dair: updated comments lemme know if this helps.

design/app-to-app.md
in2dair: Added sample data flow for photo sharing to the implementation section with both client and server steps. Let me know if I missed anything here.

design/app-to-app.md
in2dair: Very good catch. Let me add messageType to the registration. This was an oversight and it showed up as i was writing the proposed sample data flow. Let's go with 2. I don't think it precludes us from adding 3 down the road, e.g., a manifest might help the host implement disambiguation of the share target if needed (the question of whether each session gets its own registration on the host or not could be aided by manifest).

design/app-to-app.md
in2dair: i vote no for now. future message type

design/app-to-app.md
```diff
@@ -0,0 +1,111 @@
...
+### Sample data flow for photo sharing
+1. On connect (or on subsequent sharing message registration), MRE app sends a `SharingRegistration` json payload consisting of
+    * `messageType` - `string` (as Described in [API Design](#API-Design))
+2. MRE Client sends a `SharingMessage` json payload, consiting of the following fields (as Described in [API Design](#API-Design))
+    * `messageType` - `string` 
+    * `photoType` - `string`
+    * `dataSize` - `number`
+    * `transmissionMethod` - `string`, presently set to the value "websocketbinary", but may be extended, for example, to support side channel for future message types, or alternative trasnport for existing message types.
+3. MRE app invokes `SharingCallbacks.isMessageDesired` callback if present to determine whether the message is desired. If a `SharingCallback` is registered for the given `messageType` but no `isMessageDesired` is present, proceed as if the message IS desired. If no `SharingCallback` is registered for the given `messageType`, proceed as if the message is NOT desired.
+4. MRE app sends a `SharingMessageResponse` json payload, consiting of the following fields:
```
thespivey, line 104: I think you'll want some kind of id here, to correlate the responses to the original message.

design/app-to-app.md
```diff
@@ -0,0 +1,111 @@
...
+    * `result` - `boolean` true if the message is desired, false otherwise.
+5. If the message was desired, the MRE host should transmit the bytes of the image immediately following the `SharingMessageResponse` in binary format as per https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/send
```
thespivey, line 106: I'm uncertain about sending large data on the websocket.  I would tend to assume that other messages cannot be sent until the blob is fully transferred, which may cause the app to become unresponsive during a large upload?

in2dair: @thespivey  agreed on the concern, but I don't think we're boxing ourselves in here -- not much changes if we, now or down the road (let me know if you'd like me to add this now), add something like "transmissionMethod" = "secondaryWebsocketBinary", and specify that the app should open should establish a side channel (and additional web socket connection) at this stage... a few more steps but not a dramatic change.

design/app-to-app.md
in2dair: i presumed it would be the next message, but if we go the side channel route agree we should add.

design/app-to-app.md
thespivey, line 106: I think we'll want adequate header information in the blob message to indicate that this is image data for a specific SharingMessage.  If we decided to add a new blob message in the future for a different purpose, there should be no ambiguity as to what a given blob is for.

If putting a header _in_ the blob feels excessive, then consider how to build a generic mechanism into the protocol for associating blob data with a JSON message.  I'd recommend thinking about that as a transport problem, absent semantics.,   That is, you should be able as much as possible to parse the protocol and route messages with no with no knowledge of particular message types or exchanges.  This becomes important to extensibility & back-compat.

in2dair: I'll take a stab at this.

design/app-to-app.md
in2dair: added.

design/app-to-app.md
in2dair: i'll have us resend the message (in text form) on the binary side channel. lemme know if my latest commit helps.

tombuMS: Are you saying you would envision the client application needing to open a new socket on a different port to the MRE server app?

design/app-to-app.md
```diff
@@ -0,0 +1,125 @@
...
+    * `result` - `boolean` true if the message is desired, false otherwise.
+5. If the message was desired, the MRE host should establish a [Binary Side Channel](#Binary-Side-Channel) and transmit the bytes of the image immediately following the `SharingMessageResponse` in binary format.
+6. MRE app invokes `SharingCallbacks.messageReceived` with a payload similar to that of step 2, with the additional binary data of the image on the payload.
+7. MRE app sends a `SharingMessageResponse` similar to step 4, indicating whether the message was successfully received by an appropriate handler
+
+Note that future sharing message types for content smaller than images may forgo steps 4 and 5 and elect to send their entire payload with the initial json sent in step 2.
+
+
+### Binary Side Channel
+Certain message types, such as images, are not suitable for transmission on the primary websocket. In these cases a websocket bound to an additional port (TODO: Pick port number) will accept connections to handle expected payloads.
```
tombuMS, line 115: Additional ports for web sockets are not guaranteed for several hosting platforms, including Azure Web Apps.  They are sandboxed and only have ports 80 and 443 open for communication.  This seems like a less than ideal solution to this, as many of our own hosted apps are built on Azure Web App services.

in2dair: Thanks for this! Updated to restful api per your feedback!

in2dair:
Re-requesting review

tombuMS (approved):

design/app-to-app.md
in2dair: Talked offline : decided on restful approach over side websocket

PR #713, opened 2020-09-05
Add a ready promise to WebHost

stevenvergenz:
For MREs that want to add additional routes to the HTTP server, you need to wait for the server to be created before you can add your routes. This new promise tells you that the server is created and ready to go.


tombuMS (approved):
:shipit:

PR #718, opened 2020-09-22
Remove baseUrl from WebHost and tests

stevenvergenz:
`baseUrl` was only necessary because MRE hosts could not resolve relative URLs, so all assets needed to supply absolute URLs. This is no longer the case, so I'm removing `baseUrl` because it is a common source of failures for new users.

* Remove the `baseUrl` property from `WebHost`, test constructors, and asset loads.
* Remove the altspacevr-video test, as it uses a deprecated API.
* Remove custom constructor from physics-dynamic-and-kinematic test, as it's wasteful to keep assets loaded when the test isn't running.


tombuMS (changes requested):
:clock1:

tombuMS (approved):
:shipit:

patole (approved):

PR #719, opened 2020-09-24
design proposal for updated collision layer system

sorenhan:
Design proposal for changes we need to do to the collision layer to be safe for exclusive actors, as well as easier to understand and use, with more clearly defined use cases per layer

### DIFF
--- a/design/collision.layers.v2.md
+++ b/design/collision.layers.v2.md
@@ -1,0 +1,60 @@
+## One Line Concept:
+Change collision layer system to be more intuitive and safely support group masks and exclusive actors, UI, and triggers
+
+## Problem Description: 
+The collision system is used in a number of different cases, such as: Rigid Body Collision with Kinematic/Static objects or with other rigid bodies, Player Controller collision, Teleport ray collisions, UI ray collisions, proximity grab, Trigger Zones (for rigid bodies and for players.)
+
+The current collision layer implementation loosely wraps the Unity collision layers, and allowing 4 different layers (Default, NavMesh, Hologram, UI) to be assigned freely to all colliders regardless of usage.
+
+The design had to minimize number of collision layers (as that is a shared resource in Unity, and there are only 32 layers shared between all collision and all rendering), and that affected the design.
+
+The current implementation is problematic, because
+1. it's not intuitive what the different types do, or when they should be used.
+2. if using group masked/exclusive actors in certain cases, they can cause issues with rigid body simulation not being (largely) globally consistent across all connected clients
+
+I propose a new collision layer system that follow a number of requirements
+1. certain things (collisions between rigid body actors) depend on the authoritative peer client, and therefore exclusive actors and group masking cannot be allowed.
+2. anything that can happen on a specific client must be independent of the authoritative peer, to support exclusive actor and group masks - for example behaviors, personal UI, player-specific collisions and triggers
+3. each layer is configured for specific use cases and named accordingly.
+
+The system changes MRE collision layers to be more than just the equivalent of a Unity Layer, by adding additional restrictions and associated properties to each. Each layer has different requirements for
+1. Whether they can be exclusive/groupmasked, or they have to be global
+2. Which other layers it collides against
+3. Whether the collider must be triggers or non-triggers
+4. What it is generally used for
+
+I have come up with a list of 6 distinct collision layers to cover the common use cases. 
+
+ MRE Collision Layer Name|Used for|Can be exclusive/groupmasked|In Altspace uses Unity Layer|In Altspace Collides against Layers|IsTrigger in Unity
+ :--|:-:|:-:|:-:|:-:|--:
+ Physical|Rigid body, Player Collision, Blocks Cursor|no*|Default|Default, NavMesh, CharacterController, Interactables|no
+ PhysicalTeleportable|Same as Physical, but is also valid teleport destination|no*|NavMesh|Default, NavMesh, CharacterController, Interactables|no
+ CharacterBlocking|blocks charactercontroller, but not rigid bodies or cursor|yes|IgnoreRaycast|CharacterController|no
+ UI|ButtonBehavior,ToolBehavior, 2D UI|yes|MREUI(new)|None|yes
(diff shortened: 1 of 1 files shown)


design/collision.layers.v2.md
```diff
@@ -0,0 +1,60 @@
...
+4. What it is generally used for
+
+I have come up with a list of 6 distinct collision layers to cover the common use cases. 
+
+ MRE Collision Layer Name|Used for|Can be exclusive/groupmasked|In Altspace uses Unity Layer|In Altspace Collides against Layers|IsTrigger in Unity
+ :--|:-:|:-:|:-:|:-:|--:
+ Physical|Rigid body, Player Collision, Blocks Cursor|no*|Default|Default, NavMesh, CharacterController, Interactables|no
+ PhysicalTeleportable|Same as Physical, but is also valid teleport destination|no*|NavMesh|Default, NavMesh, CharacterController, Interactables|no
+ CharacterBlocking|blocks charactercontroller, but not rigid bodies or cursor|yes|IgnoreRaycast|CharacterController|no
+ UI|ButtonBehavior,ToolBehavior, 2D UI|yes|MREUI(new)|None|yes
```
stevenvergenz, line 33: >MREUI(new [](start = 43, length = 9)

Is there a reason this needs to be a new layer? Why not use the existing UI layer?

design/collision.layers.v2.md
```diff
@@ -0,0 +1,60 @@
...
+1. Whether they can be exclusive/groupmasked, or they have to be global
+2. Which other layers it collides against
+3. Whether the collider must be triggers or non-triggers
+4. What it is generally used for
+
+I have come up with a list of 6 distinct collision layers to cover the common use cases. 
+
+ MRE Collision Layer Name|Used for|Can be exclusive/groupmasked|In Altspace uses Unity Layer|In Altspace Collides against Layers|IsTrigger in Unity
+ :--|:-:|:-:|:-:|:-:|--:
+ Physical|Rigid body, Player Collision, Blocks Cursor|no*|Default|Default, NavMesh, CharacterController, Interactables|no
```
stevenvergenz, line 30: >CharacterController [](start = 84, length = 19)

Should physics interact with the player? Our player colliders are not very tight, it could be a weird experience.

design/collision.layers.v2.md
```diff
@@ -0,0 +1,60 @@
...
+ PhysicalTeleportable|Same as Physical, but is also valid teleport destination|no*|NavMesh|Default, NavMesh, CharacterController, Interactables|no
+ CharacterBlocking|blocks charactercontroller, but not rigid bodies or cursor|yes|IgnoreRaycast|CharacterController|no
```
stevenvergenz, line 32: >CharacterBlocking [](start = 1, length = 17)

Needs to block teleport rays too, which is a subset of the cursor. Not sure how to tease those out.

design/collision.layers.v2.md
```diff
@@ -0,0 +1,60 @@
...
+## Problem Description: 
+The collision system is used in a number of different cases, such as: Rigid Body Collision with Kinematic/Static objects or with other rigid bodies, Player Controller collision, Teleport ray collisions, UI ray collisions, proximity grab, Trigger Zones (for rigid bodies and for players.)
+
+The current collision layer implementation loosely wraps the Unity collision layers, and allowing 4 different layers (Default, NavMesh, Hologram, UI) to be assigned freely to all colliders regardless of usage.
+
+The design had to minimize number of collision layers (as that is a shared resource in Unity, and there are only 32 layers shared between all collision and all rendering), and that affected the design.
+
+The current implementation is problematic, because
+1. it's not intuitive what the different types do, or when they should be used.
+2. if using group masked/exclusive actors in certain cases, they can cause issues with rigid body simulation not being (largely) globally consistent across all connected clients
```
jabenk, line 13: I do not see why this should be an issue. I think the collision layers are syncronized across all clients then even they are not globally consistent but they always should tend to the same global state. 
Or is it meant here that the prediction code does not support (yet) the different collision layers (?)

design/collision.layers.v2.md
```diff
@@ -0,0 +1,60 @@
...
+
+ MRE Collision Layer Name|Used for|Can be exclusive/groupmasked|In Altspace uses Unity Layer|In Altspace Collides against Layers|IsTrigger in Unity
+ :--|:-:|:-:|:-:|:-:|--:
+ Physical|Rigid body, Player Collision, Blocks Cursor|no*|Default|Default, NavMesh, CharacterController, Interactables|no
+ PhysicalTeleportable|Same as Physical, but is also valid teleport destination|no*|NavMesh|Default, NavMesh, CharacterController, Interactables|no
+ CharacterBlocking|blocks charactercontroller, but not rigid bodies or cursor|yes|IgnoreRaycast|CharacterController|no
+ UI|ButtonBehavior,ToolBehavior, 2D UI|yes|MREUI(new)|None|yes
+ PersonalTrigger|Trigger Volumes for charactercontrollers and exclusive (not group masked) rigid bodies|yes|MREPersonalTrigger(new)|Default,NavMesh,CharacterController|yes
+ GlobalTrigger|Trigger Volumes for global (not group masked/exclusive) rigid bodies|no|MREGlobalTrigger(new)|Default,NavMesh,CharacterController|yes
+
```
jabenk, line 36: Probably I do not know the use case but for me, it sounds like physical and physicalteleportable could be just one.

design/collision.layers.v2.md
```diff
@@ -0,0 +1,60 @@
...
+2. We should warn when button (possibly certain tool) behavior is using layer other than UI
+
+## Implementation order
+1. add new layers to sdk and runtime, and configure testbeds and altspace implementations, including the collision masks
+2. remove istrigger flag from MRE API and protocol, and drive it from Unity Runtime
+3. enforce errors on actor creation
+4. enforce errors on collision layer change after actor creation
+5. check for warning on button behavior assignment
+6. Make triggers filter out the events that don't match the requirements
+
```
jabenk, line 60: 1) should add this layer support to the prediction code, as far as I know, the jitter buffer should not know about the collision layers. 
If not added would not be a show stopper but could eliminate unnecessary artifacts.

PR #725, opened 2020-10-02
button-behavior test V0

patole:
This test has just a square button in the middle of the scene and text above it. The text changes from the default state to indicate if the button detects either a quick click or a hold, then back to the default text after 1000 ms.

### DIFF
--- a/packages/functional-tests/src/tests/button-behavior-test.ts
+++ b/packages/functional-tests/src/tests/button-behavior-test.ts
@@ -1,0 +1,90 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import * as MRE from '@microsoft/mixed-reality-extension-sdk';
+
+import { Test } from '../test';
+
+export default class ButtonBehaviorTest extends Test {
+	private defaultLabel = "Test the difference between click and hold";
+	public expectedResultDescription = "Button behavior";
+
+	private testButton: MRE.Actor;
+	private testBehavior: MRE.ButtonBehavior;
+	private buttonLabel: MRE.Actor;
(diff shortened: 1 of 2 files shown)


packages/functional-tests/src/tests/button-behavior-test.ts
```diff
@@ -0,0 +1,90 @@
...
+					height: .1,
+					anchor: MRE.TextAnchorLocation.BottomCenter,
+					color: MRE.Color3.Teal()
+				}
+			}
+		})
+
+		const testButtonBehavior = this.testButton.setBehavior(MRE.ButtonBehavior);
+		testButtonBehavior.onClick((_, __) => this.displayString("Click Detected!"));
+		testButtonBehavior.onButton('holding', () => this.displayString("Hold Detected!"));
```
tombuMS, line 88: >'holding' [](start = 30, length = 9)

I see this is written as v0 in the title.  Is the desire to expand this to cover and test the other events for the button?

patole: Good question! I wrote this based on a sentence or two description from @sorenhan and wanted his input to see if there were more events desired in this, or if simply a click/press threshold test was sufficient.

tombuMS: Well as the behavior system area owner, I would totally love to expand this and fully test the button.  Let me know when you have time to discuss expanding this if you wanted to take that on as well.  :)

tombuMS (approved):
:shipit:

packages/functional-tests/src/tests/button-behavior-test.ts
patole: Sure thing, we can write this up after you're done with the media player work so as not to distract from that.

PR #729, opened 2020-10-13
[gltf-gen] Import existing gltf/glb files into a factory for editing

stevenvergenz:
* Add images stored in buffers as a valid texture source
* Fix some export de-duplication bugs in mesh, node, and scene
* Add `importFromGltf` and `importFromGlb` methods to the `GltfFactory` class. These parse out the glTF data and convert it into the GltfGen representation for editing. Once it's converted and stored in the factory, you can make changes to anything you want, then re-export.


tombuMS (approved):
:shipit:

PR #737, opened 2020-12-31
Don't sync transforms of grabbed exclusive actors

stevenvergenz:
This fixes a bug that causes actor creation lockups on clients that receive these updates.


patole (commented):
:shipit:

Issue #739, opened 2021-03-12
Transform updates not propagating for attachments when using subscription 'transform'


David032:
It was flagged in the MRE-SDK discord [here](https://discordapp.com/channels/560149912231346177/560516951437344778/818625109052096552) that transform updates do not propagate to attachments when using the subscription 'transform'. A fix has been found but as it requires either the modification of the built code in the node_modules folder (Lines 186 through to the next '{' of rules.js) or the sdk itself (403-406 of rules.ts) it's not viable for most deployments.

The fix:
`const syncActor = session.actorSet.get(message.payload.actor.id);
 const attachment = syncActor.initialization.message.payload.actor.attachment;
 if (syncActor && ((client.authoritative && !syncActor.grabbedBy) ||
    (syncActor.grabbedBy === client.id) || 
    (attachment && attachment.userId === client.userId))) { `

amoringello:
The fix works locally but like stated, that is not a viable plan for deployment. It needs to be in the SDK.  I'm stuck running on localhost-only where I can make the changes myself.  This needs to

PR #740, opened 2021-03-12
Transform updates not propagating for attachments when using subscription 'transform'

David032:
It was flagged in the MRE-SDK discord [here](https://discordapp.com/channels/560149912231346177/560516951437344778/818625109052096552) that transform updates do not propagate to attachments when using the subscription 'transform' 
This could be intentional, I don't know? But if it's not, this fix & the attached issue #739 corrects it


tombuMS:
Pull request #745 addresses this fix from both client -> app and app -> client.  Closing in favor of that fix once it is ready.

Issue #742, opened 2021-03-23
Rigidbody Addforce or Velocity no longer working?


AlecGamble:
Hi,

I've recently been developing a minigolf SDK and as of the update this week my Rigidbody physics which I was applying to the ball no longer works. I can no longer AddForce or set the velocity as I was doing before. logging the rigidbody's velocity length after using AddForce prints 0s as if it hasn't been applied and setting the velocity manually returns the length of the vector it has been set to so it has been applied but has no visible effect in-game. I've tried creating a new project to isolate this behaviour but I'm still having no luck applying a force to it. This is a behaviour I had working before and initially I thought it might have been that because I was updating my kit objects last week it was using a cached version and the new version had something different but having tried to isolate the behaviour in the following script I think it might be a bug.

```typescript
export default class App {
    private assets: MRE.AssetContainer;
    private ball: MRE.Actor;
    private speedCheckTimer: NodeJS.Timer;

    constructor(private context: MRE.Context) 
    {
        this.context.onStarted(() => this.Initialise());
    }

    private Initialise() 
    {
        this.assets = new MRE.AssetContainer(this.context);
        this.ball = MRE.Actor.CreatePrimitive(this.assets, {
            definition: {
                shape: MRE.PrimitiveShape.Sphere,
                dimensions: { x: 0.2, y: 0.2, z: 0.2 }
            },
            addCollider: true,
            actor: {
                name: 'ball prim',
                transform: { local: { position: { x: 0, y: 0, z: 0 } } },
                rigidBody: {
                    useGravity:true,
                    mass: 0.2,
                    // velocity: MRE.Vector3.Forward().scale(50000) // <--------
                },
                subscriptions: ['transform', 'rigidbody', 'rigidbody-velocity'],

            }
        });
        this.ball.collider.layer = MRE.CollisionLayer.Navigation;
        // this.ball.rigidBody.velocity = MRE.Vector3.Forward().scale(50000); // <--------
        // this.ball.rigidBody.addForce(MRE.Vector3.Forward().scale(50000)); // <--------
        this.CheckBallSpeed();
    }

    private CheckBallSpeed()
    {
        let x = 0;
        this.speedCheckTimer = setInterval(() => {
            let currentVelocity = this.ball.rigidBody.velocity;
            console.log(currentVelocity.length());
            x++;
            if(x > 5) clearInterval(this.speedCheckTimer);
        }, 120);
    }
}
```

Thanks in advance for your help!

[...]

PR #743, opened 2021-03-25
Actors with orphaned ownership will properly update upon new authoritative user

norybiak:
From my testing, `_rigidBodyDefaultOwner` has already changed to the new authoritative user before `userLeft()` is called. This means that the actors owned by the user that just left changes ownership to the new authoritative user without the actor being orphaned. 

However, if  an actor were orphaned, it wouldn't properly update ownership when a new authoritative user is assigned. This fix ensures that ownership is properly changed.

Related issue #742

This PR is required for PR #744


tombuMS (approved):
🚢

PR #745, opened 2021-03-25
Fix for transform subscriptions when actor is an attachment

norybiak:
This fix allows transform subscriptions to pass to the app when actor is an attachment.

Issue #739

This PR is similar to #740 but for the correct branch (red).

### DIFF
--- a/packages/sdk/src/internal/adapters/multipeer/rules.ts
+++ b/packages/sdk/src/internal/adapters/multipeer/rules.ts
@@ -305,8 +305,10 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
 				message: Message<Payloads.ActorCorrection>
 			) => {
 				const syncActor = session.actorSet.get(message.payload.actorId);
-				if (syncActor && ((client.authoritative && !syncActor.grabbedBy)
-					|| (syncActor.grabbedBy === client.id))) {
+				const attachment = syncActor.initialization.message.payload.actor.attachment;
+				if (syncActor && ((client.authoritative && !syncActor.grabbedBy) ||
+					(syncActor.grabbedBy === client.id) || 
+                    (attachment && attachment.userId === client.userId))) {
 					const correctionPayload = message.payload;
 
 					// Synthesize an actor update message and add in the transform from the correction payload.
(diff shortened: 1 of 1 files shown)


packages/sdk/src/internal/adapters/multipeer/rules.ts
```diff
@@ -305,8 +305,10 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
 				message: Message<Payloads.ActorCorrection>
 			) => {
 				const syncActor = session.actorSet.get(message.payload.actorId);
-				if (syncActor && ((client.authoritative && !syncActor.grabbedBy)
-					|| (syncActor.grabbedBy === client.id))) {
+				const attachment = syncActor.initialization.message.payload.actor.attachment;
```
tombuMS: It is possible for syncActor to be undefined here which will cause this to throw an exception.  Getting the attachment will need to be moved into the if block below it where syncActor will already be tested for undefined.  Alternatively you can do optional chaining on the syncActor like: 
```ts
const attachment = syncActor?.initialization.message.payload.actor.attachment;
```

packages/sdk/src/internal/adapters/multipeer/rules.ts
```diff
@@ -402,8 +404,10 @@ export const Rules: { [id in Payloads.PayloadType]: Rule } = {
 				message: Message<Payloads.ActorUpdate>
 			) => {
 				const syncActor = session.actorSet.get(message.payload.actor.id);
+				const attachment = syncActor.initialization.message.payload.actor.attachment;
```
tombuMS: See same comment above about syncActyor being undefined.

tombuMS (changes requested):
Some protection regarding potentially undefined syncActor needs to be added.  See comments

packages/sdk/src/internal/adapters/multipeer/rules.ts
norybiak: Fixed. See latest commit.

packages/sdk/src/internal/adapters/multipeer/rules.ts
norybiak: Fixed. See latest commit.

tombuMS (approved):

PR #749, opened 2021-08-04
Fixes several memory leaks associated with destroying an actor

kuperavv:
Summary of changes:

Creates a public method on actor that unobserves the collider and breaks the reference to the collider.
Removes mesh references to the actor to be destroyed
Removes material references to the actor to be destroyed
Added the ability to set the Collider layer using setCollider.
    - Setting the collider when instantiating an actor creates a memory leak due the manner in which the collider is attached to the actor.  Left a TODO to resolve this, but did not want to introduce any code breaking changes.

Unobserves 'other' collider listeners when copying to a new collider.

### DIFF
--- a/packages/sdk/src/actor/actor.ts
+++ b/packages/sdk/src/actor/actor.ts
@@ -502,15 +502,12 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 		center = { x: 0, y: 0, z: 0 } as Vector3Like,
 		layer = CollisionLayer.Default,
 	): void {
-		const colliderGeometry = this.generateColliderGeometry(colliderType, size, center);
-		if (colliderGeometry) {
-			this._setCollider({
-				enabled: true,
-				isTrigger,
-				layer,
-				geometry: colliderGeometry
-			} as ColliderLike);
-		}
+		this._setCollider({
+			enabled: true,
+			isTrigger,
+			layer,
+			geometry: { shape: colliderType, size, center}
+		} as ColliderLike);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


packages/sdk/src/actor/actor.ts
```diff
@@ -198,6 +198,16 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 	 * PUBLIC METHODS
 	 */
 
+	/**
+	 * Destroys the collider.
+	 */
+	public clearCollider(): void {
```
stevenvergenz, line 204: Good find! I don't think I've ever needed to remove a collider without removing the whole actor before, but I agree it should be possible.

packages/sdk/src/core/contextInternal.ts
```diff
@@ -450,10 +450,13 @@ export class ContextInternal {
 			const isNewActor = !this.actorSet.get(sactor.id);
 			const actor = isNewActor ? Actor.alloc(this.context, sactor.id) : this.actorSet.get(sactor.id);
 			this.actorSet.set(sactor.id, actor);
+
+			//TODO - Fix memory leak that occurs when instantiating an actor with a collider. Use setCollider instead.
+			//Collider is attached to actor via actor.copy(sactor), but it creates a memory leak.
```
stevenvergenz: `actor.copy` calls `actor._setCollider` just like `actor.setCollider` does, which appears to unobserve the old collider before rebinding. Is it the collider or the actor that's leaking? Does this still apply with your call to `clearCollider` on local destroy?

stevenvergenz (commented):
Great first PR!

packages/sdk/src/core/contextInternal.ts
kuperavv: It's the collider that is leaking during instantiation.  It's about ~50% of the leakage I've detected actually.  If I don't set it during instantiation, I don't observe the memory leak.  And yes, it is separate from the clearCollider cleanup leakage, which was maybe ~10%.  I'll take another look at actor.copy and see if I can identify the source.  I'll admit that the issue there isn't immediately apparent to me.

kuperavv:
I was able to identify and resolve the memory leak associated with calling _setCollider() from within actor.copy().  

There was an existing reference to the collider through the collider geometry.

To resolve this, and to only call generateColliderGeometry once, I deferred generateColliderGeometry in setCollider() into _setCollider().  Due to the typing, I added an if/then/else block to re-assign size and center.  I then call generateColliderGeometry to create a new geometry, and re-assign it to the collider to break the references.

I removed the TODO comments as now you can instantiate an actor simultaneously with the collider without the memory leak.

kuperavv:
I resolved some additional memory leaks stemming from not cleaning up animation references.

tombuMS:
Thanks for the first PR!

This PR is targeting the wrong branch though.  Can you please retarget it to the `red` branch for further review?

kuperavv:
@tombuMS - Thanks for pointing that out.  I believe I should have submitted another PR targeting the red branch, can you confirm that looks correct on your end?  We can close this PR request if it looks correct.

PR #750, opened 2021-08-10
targeting red branch

kuperavv:
Summary of changes:

-Removes animation references of the actor to be destroyed.

-Creates a public method on actor that destroys a collider (unobserves the collider and breaks the reference to the collider).

-Removes the collider of the actor to be destroyed.

-Removes mesh references of the actor to be destroyed.

-Removes material references of the actor to be destroyed.

-Added the ability to set the Collider layer using setCollider.

-Resolves a memory leak associated with creating an actor containing a collider, via actor.copy(). Setting the collider via setCollider() does not create a memory leak. To avoid calling generateColliderGeometry multiple times, I deferred the call of generateColliderGeometry() in setCollider into _setCollider(). This creates a fresh collider geometry whether the collider is created via setCollider() or during an actor instantiation, via actor.copy(). The fresh collider geometry is then re-attached to collider, but breaks the references causing the memory leak.

-Unobserves 'other' collider listeners when copying to a new collider.

### DIFF
--- a/packages/sdk/src/actor/actor.ts
+++ b/packages/sdk/src/actor/actor.ts
@@ -432,7 +432,6 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 	 * based on the size of the currently assigned mesh (loading meshes are not considered).
 	 * If no mesh is assigned, defaults to 0.5.
 	 * @param center The center of the collider, or default of the object if none is provided.
-	 * @param layer Controls what the assigned actors will collide with.
 	 */
 	// * @param collisionLayer The layer that the collider operates in.
 	public setCollider(
@@ -441,7 +440,6 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 		isTrigger: boolean,
 		radius?: number,
 		center?: Vector3Like,
-		layer?: CollisionLayer
 	): void;
 
 	/**
@@ -452,14 +450,12 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 	 * based on the currently assigned mesh (loading meshes are not considered).
 	 * If no mesh is assigned, defaults to (1,1,1).
 	 * @param center The center of the collider, or default of the object if none is provided.
-	 * @param layer Controls what the assigned actors will collide with.
 	 */
 	public setCollider(
 		colliderType: ColliderType.Box,
 		isTrigger: boolean,
 		size?: Vector3Like,
 		center?: Vector3Like,
-		layer?: CollisionLayer
 	): void;
 
 	/**
@@ -471,30 +467,22 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 	 * If omitted, a best-guess size is chosen based on the currently assigned mesh
 	 * (loading meshes are not considered). If no mesh is assigned, defaults to (1, 1, 1).
 	 * @param center The center of the collider, or default of the object if none is provided.
-	 * @param layer Controls what the assigned actors will collide with.
 	 */
 	public setCollider(
 		colliderType: ColliderType.Capsule,
 		isTrigger: boolean,
 		size?: Vector3Like,
 		center?: Vector3Like,
-		layer?: CollisionLayer,
 	): void;
 
 	/**
 	 * Adds a collider whose shape is determined by the current mesh.
 	 * @param colliderType Type of the collider to enable.
 	 * @param isTrigger Whether the collider is a trigger volume or not.
-	 * @param size Ignored for ColliderType Auto.
-	 * @param center Ignored for ColliderType Auto.
-	 * @param layer Controls what the assigned actors will collide with.
 	 */
 	public setCollider(
 		colliderType: ColliderType.Auto,
 		isTrigger: boolean,
-		size?: Vector3Like,
-		center?: Vector3Like,
-		layer?: CollisionLayer
 	): void;
 
 	public setCollider(
--- a/packages/sdk/src/actor/actor.ts
+++ b/packages/sdk/src/actor/actor.ts
@@ -490,12 +490,10 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 		isTrigger: boolean,
 		size?: number | Vector3Like,
 		center = { x: 0, y: 0, z: 0 } as Vector3Like,
-		layer = CollisionLayer.Default,
 	): void {
 		this._setCollider({
 			enabled: true,
 			isTrigger,
-			layer,
 			geometry: {shape: colliderType, size, center}
 		} as ColliderLike);
 	}
--- a/packages/sdk/src/actor/actor.ts
+++ b/packages/sdk/src/actor/actor.ts
@@ -199,6 +199,16 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 	 */
 
 	/**
+	* Destroys the collider.
+	*/
+	public clearCollider(): void {
+		if (this._collider) {
+			unobserve(this._collider);
+			this._collider = null;
+		}
+	}
+
+	/**
 	 * Creates a new, empty actor without geometry.
 	 * @param context The SDK context object.
 	 * @param options.actor The initial state of the actor.
@@ -422,6 +432,7 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
 	 * based on the size of the currently assigned mesh (loading meshes are not considered).
 	 * If no mesh is assigned, defaults to 0.5.
 	 * @param center The center of the collider, or default of the object if none is provided.
+	 * @param layer Controls what the assigned actors will collide with.
 	 */
 	// * @param collisionLayer The layer that the collider operates in.
 	public setCollider(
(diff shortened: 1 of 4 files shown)


tombuMS:
Thanks for retargeting to red!  I will have a look at this within the next day or two.  Thanks again for the PR!

kuperavv:
Thank you!  I closed the other PR.

kuperavv:
@tombuMS any feedback on this PR?

jgwinner:
Feedback would be good, I'm about to shove 2,000 people through an MRE :)

packages/sdk/src/core/contextInternal.ts
```diff
@@ -627,6 +627,29 @@ export class ContextInternal {
 			this.localDestroyActor(child);
 		});
 
+		//Remove animations
+		for (const anim of actor.targetingAnimations.values()) {
+			anim.data.clearReference(anim)
+			anim.data.breakReference(anim);
+		}
```
tombuMS, line 633: The reference to the animation here should not be broken based on an actor being destroyed.  The animation data can contain more than one target, and they do not all need to be valid targets for the animation to work and still be valid.  When the actor is deleted from the actor set below this new block of code, the animation will get cleaned up a few lines later where we check to see if it is an orphaned animation.

packages/sdk/src/actor/actor.ts
```diff
@@ -838,24 +856,44 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+		} else if (collider.geometry.shape === ColliderType.Capsule) {
+			size = collider.geometry.size;
+			center = collider.geometry.center;
 		}
 
-		this._collider = new Collider(this, collider);
-		if (oldCollider) {
-			this._collider.internal.copyHandlers(oldCollider.internal);
-		}
+		const geometry = this.generateColliderGeometry(collider.geometry.shape, size, center)
```
tombuMS, line 859: What is the purpose of moving the collider geometry unpacking to here rather than in the main setCollider function?

packages/sdk/src/actor/actor.ts
```diff
@@ -458,40 +471,45 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
+	 * @param size Ignored for ColliderType Auto.
+	 * @param center Ignored for ColliderType Auto.
+	 * @param layer Controls what the assigned actors will collide with.
 	 */
 	public setCollider(
 		colliderType: ColliderType.Auto,
-		isTrigger: boolean
+		isTrigger: boolean,
+		size?: Vector3Like,
+		center?: Vector3Like,
```
tombuMS: These parameters do not make sense at an API level for a collider of type auto.  I am guessing this is to support the change of moving the collider geometry determination to the _setCollider function?  Please see my question there about the reasons you are wanting to make this change.

packages/sdk/src/actor/actor.ts
```diff
@@ -458,40 +471,45 @@ export class Actor implements ActorLike, Patchable<ActorLike> {
...
 	 * @param center The center of the collider, or default of the object if none is provided.
+	 * @param layer Controls what the assigned actors will collide with.
 	 */
 	public setCollider(
 		colliderType: ColliderType.Capsule,
 		isTrigger: boolean,
 		size?: Vector3Like,
-		center?: Vector3Like
+		center?: Vector3Like,
+		layer?: CollisionLayer,
```
tombuMS: Collision layers was disabled here for a specific reason as we do not support changing the collision layer after the actor has been created.  If there is a need for this to meet some kind of functional reuirement, then please pull this out of this PR and submit a feature request PR with the change so that we can discuss it there.

packages/sdk/src/actor/physics/colliderInternal.ts
```diff
@@ -49,6 +49,9 @@ export class ColliderInternal {
 		for (const event of other._eventHandlers.eventNames()) {
 			for (const handler of other._eventHandlers.listeners(event)) {
 				this._eventHandlers.on(event, handler as (...args: any[]) => void);
+
+				//Remove other event handlers during copy
+				other._eventHandlers.off(event, handler as (...args: any[]) => void)
```
tombuMS, line 54: Good catch.  Thanks!

packages/sdk/src/core/contextInternal.ts
```diff
@@ -627,6 +627,29 @@ export class ContextInternal {
...
+		if (actor.appearance.mesh) {
+			actor.appearance.mesh.clearReference(actor)
+			actor.appearance.mesh.breakReference(actor);
+		}
+
+		//Remove material reference
+		if (actor.appearance.material) {
+			actor.appearance.material.clearReference(actor)
+			actor.appearance.material.breakReference(actor);
+		}
```
tombuMS, line 650: Nice work fixing these three here!

tombuMS (changes requested):
Overall nice fixes!  Couple of bits of feedback and questions.  It's definitely close to pulling in!

tombuMS:
> Feedback would be good, I'm about to shove 2,000 people through an MRE :)

Feedback given, but you should know that this won't go out for a release in to Altspace for a bit.  There are some release blockers before MRE can release the next version in to Altspace, so I hope you aren't counting on specific fixes in here in the very near future.

packages/sdk/src/core/contextInternal.ts
kuperavv: Yes, agreed after a second look.  I still detected a (small) memory leak associated destroying actors with animation data.  Maybe it was GUID in the _targetingIds.  I'll add a method to remove that id and see what you think.

packages/sdk/src/actor/actor.ts
kuperavv: I agree.

packages/sdk/src/actor/actor.ts
kuperavv: I'll revert this back.

packages/sdk/src/actor/actor.ts
kuperavv: The most significant memory leak occurs when a collider is created when instantiating an actor (via actor.copy()).  However, I observed that the memory leak does not occur when you create the collider after creating an actor via setCollider.  The only difference between those methods is the unpacking of the collider geometry (namely, you don't do that when you create the collider via actor.copy().  I deferred the generateColliderGeometry to ensure the geometry gets unpacked regardless if the collider gets created via setCollider or actor.copy().  I no longer observe the memory leak after this modification.

kuperavv:
@tombuMS - I agree with your feedback on the animation and the collider layer (and reverted back to the original base code).   Please see the comments on the reasoning for proposed changes on _generateColliderGeometry to avoid creating the memory leak the occurs when creating a collider during an actor instantiation (via actor.copy()).

packages/sdk/src/core/contextInternal.ts
tombuMS: Looks good.  Thanks!

packages/sdk/src/actor/actor.ts
tombuMS: I think that we can clean this up a bit from an API and code perspective, as there is kind of a double branching unpack here with regards to the size.  What if we updated the `_setCollider` function to take an optional `geometryDescription` parameter of type:

```ts
export type GeometryDescription = {
	shape: ColliderType;
	center?: Readonly<Vector3Like>;
	size?: Readonly<Vector3Like> | number;
};
```

Then `setCollider` can pass in the geometry description for `_setCollider` to use instead of abusing the fact that geometry is a merged type, and use the description object to call `generateColliderGeometry` to do the unpacking without the added branching in this function.  Actor copy can then just pass in the collider itself and we would use that directly to copy in to the new collider.  It should eliminate the extra branching in this function and rely on a more explict API rather than do some magic thanks to JavaScript and the TS merged types.

Thoughts?

kuperavv: @tombuMS Yes, I don't love the extra branching as it stands now. The extra branching is actually to unpack the collider that is coming in via actor.copy().  I can't access center and size properties of the collider.geometry until it's type is explicitly known.

Because generateColliderGeometry takes params (colliderType, size, center), I can't simply pass in the geometryDescription; it needs to be destructured  first (requiring access to size and center properties).  To access them, I still need to some type of:
let shape;
if (collider.geometry.shape) {
 shape = collider.geometry.shape
}

Thoughts?  I may have misunderstood your thoughts as well.

tombuMS: @kuperavv any update with this?  We are going to be working toward an MRE release and would love to include this in that release.

kuperavv: @tombuMS  - sorry about the pause in the action.  I accidentally left the post (above) in draft mode, and never sent it.  It was a few months ago, but I recall trying your suggestion, but I couldn't get it to pass through the TS checks.

kuperavv: @tombuMS  - any additional feedback or suggestions on this?

